PR #2, opened 2021-07-28
Add some notes on the design process

nrc:
r? @rylev Covers our discussion yesterday and some things from Zulip.

### DIFF
--- a/surveys/notes/README.md
+++ b/surveys/notes/README.md
@@ -1,0 +1,3 @@
+# Notes
+
+We would like this repo to serve as a template for producing future Rust surveys. Towards that goal we will record various parts of the process in this directory with the intention of transforming these notes into some kind of guide or user manual once we're clear on what the best format would be.
--- a/surveys/notes/process.md
+++ b/surveys/notes/process.md
@@ -1,0 +1,19 @@
+# Process
+
+See [2021-annual-survey#tentative-timeline](https://github.com/rust-lang/surveys/tree/main/surveys/2021-annual-survey#tentative-timeline) for the 2021 timeline.
+
+## Design doc
+
+See also [design.md](design.md).
+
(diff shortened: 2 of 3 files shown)


surveys/notes/README.md
```diff
@@ -0,0 +1,3 @@
+# Notes
```
rylev, line 1: Should this whole section be called `docs` or `process` instead?

surveys/notes/process.md
```diff
@@ -0,0 +1,31 @@
...
+See also [design.md](design.md).
+
+We intend to start the survey design process by writing a design document. Starting with the 'requirements': the lessons we hope to learn from the survey, we will work backwards to the data we hope to collect (in broad terms and including the cohorts we may wish to analyze), and finally to the questions we wish to ask in the survey.
+
+
+## Requirement collection
+
+To ensure we meet the requirements of 'clients' of the survey (i.e., people who will consume the processed result data), we intend to ask the following groups of people for input:
+
+TODO tracking issue?
```
rylev, line 26: I can create this issue.

surveys/notes/process.md
```diff
@@ -0,0 +1,31 @@
+# Process
+
+Proposed timeline:
```
rylev, line 3: This timeline is already included here: https://github.com/rust-lang/surveys/tree/main/surveys/2021-annual-survey#tentative-timeline

Not sure we need both.

rylev (approved):
Looking good! I had one question about the inclusion of the timeline, but there's no need to block merging this.

surveys/notes/README.md
nrc: I thought we might make notes on other things too, so have kept process to one file. I thought 'notes' better reflects the quality of the documentation than 'docs' :-)

PR #18, opened 2021-09-03
Add some questions from previous years

nrc:
Most of these I just think we should remove.

### DIFF
--- a/surveys/2021-annual-survey/questions.md
+++ b/surveys/2021-annual-survey/questions.md
@@ -4,7 +4,7 @@
 
 Select one:
 
-- Yes, I have used for Rust in the past three months
+- Yes, I have used Rust in the past three months
 - I have used Rust in the past, but not in the past three months
 - No, I have never used Rust
 
@@ -23,7 +23,9 @@ Select one:
 
 > **REMOVED**
 >
-> Not useful
+> Not useful. Possible uses I see are cohort analysis (did people who stopped quickly have different
+> reasons compared to those who took longer?), or tracking over time whether people spend longer
+> with Rust. However, I don't see either of these data being actionable.
 
 ### How long ago did you stop using Rust:
 
@@ -31,7 +33,7 @@ Select one:
 
 > **REMOVED**
 >
-> Not useful
+> Not useful, same reasoning as above.
 
 ### Why did you stop using Rust?
 
@@ -52,6 +54,9 @@ Select all that apply:
 > **justification**
 >
 > Useful for understanding why people give up on Rust.
+>
+> TODO should we update the answer list? I might merge the tools and IDE options, and reword
+> "Rust didn't solve a problem for me" to "I did not see any benefit to using Rust" or something similar.
 
 ### If you indicated a lack of support for platforms or architectures that you would like to target, which would those be?
 
@@ -59,7 +64,8 @@ Free form.
 
 > **REMOVED**
 >
-> Vague, not useful
+> Vague, not useful (platform support is primarily driven by supply rather than demand, there are
+> unlikely to be platforms which are unknown to us and in demand).
 
 ### If you indicated a desire for increased interoperability, which of the following languages would you want to use with Rust?
 
@@ -67,7 +73,7 @@ Select all that apply:
 
 > **REMOVED**
 >
-> Not useful
+> Not useful (support for language interop is community, rather than project, driven).
 
 ### Please provide any additional details on why you stopped using Rust.
 
@@ -75,7 +81,7 @@ Free form.
 
 > **REMOVED**
 >
-> Vague, not useful
+> Vague.
 
 
 ## For non-Rust users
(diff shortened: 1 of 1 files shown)


surveys/2021-annual-survey/questions.md
```diff
@@ -1,19 +1,138 @@
 # Survey questions
 
-## About you
+### Do you use Rust?
 
-See [who](./design/who.md).
+Select one:
 
-The following are primarily for cohort analysis, secondarily for understanding the shape of the community.
+- Yes, I have used for Rust in the past three months
```
rylev, line 7: We also likely want to base control flow of the survey on this question. Not sure yet how to capture that.

surveys/2021-annual-survey/questions.md
```diff
@@ -1,19 +1,138 @@
...
-- I have never used Rust
+## For previous Rust users
+
+### How long did you use Rust before you stopped?
+
+Select one:
+
+> **REMOVED**
+>
+> Not useful
```
rylev, line 26: Can we explain a bit more *why* it's not useful?

surveys/2021-annual-survey/questions.md
```diff
@@ -1,19 +1,138 @@
...
+
+### Why did you stop using Rust?
+
+Select all that apply:
+
+- Rust didn't solve a problem for me
```
rylev, line 40: I really don't like this answer as I'm not sure what it actually means...

rylev: In general I'd like to think through whether the group of answers we have here is the right one.

surveys/2021-annual-survey/questions.md
```diff
@@ -1,19 +1,138 @@
...
+- My company doesn't use Rust
+- Switching to Rust would slow me down too much
+- I haven't learned Rust yet, but I want to
+- I haven't learned Rust yet, and I don't want to
+- Other
+
+> **justification**
+>
+> Useful for understanding why people don't use Rust without trying, even though they are interested enough to complete this survey.
+> TODO Given the strong bias in the sample, I'm not sure if this data is actually useful.
```
rylev, line 104: I wonder if there's a way to better understand why someone has never used Rust but is filling out this survey. Perhaps that can debias this data a bit? Overall, I'm also not sure about this question.

surveys/2021-annual-survey/questions.md
```diff
@@ -288,3 +407,103 @@ Is Rust mandated for your course or activity, or did you choose it yourself?
...
+>
+> Vague, not useful.
+
+### If you are interested in attending a Rust conference, which of these regions would you travel to?
+
+Select all that apply
+
+> **REMOVED**
+>
+> Not useful.
```
rylev, line 446: Are we sure this is not useful? For regional conferences, this might give them an indication of how likely their conference is to draw a crowd from outside their region. For non-regional conferences (e.g., RustFest) this can give them an indicator of additional places they may want to look into hosting. 

Overall, I can see why you'd think this isn't useful, but perhaps we should ship it around to conference organizers to get their input.

surveys/2021-annual-survey/questions.md
```diff
@@ -288,3 +407,103 @@ Is Rust mandated for your course or activity, or did you choose it yourself?
...
+
+## Contact information
+
+### Email address
+
+Free form
+
+> **REMOVED**
+>
+> We aren't going to contact anyone
```
rylev, line 491: If this is the case, I think maybe we should write more about why. Obviously someone thought it was useful when they first created the survey

surveys/2021-annual-survey/questions.md
nrc: Yeah, I was wondering about that. I'm not a great fan of the current set, but I figure they were good enough and it would be good to have continuity for this question

surveys/2021-annual-survey/questions.md
nrc: From previous conversations, most conference organisers start with a location, usually local, rather than thinking "I want to put in a conference, where should I do it". I think this would only be useful if the foundation or maybe rustfest wanted to do another conference

surveys/2021-annual-survey/questions.md
nrc: Perhaps I'm being too cynical, but I can't imagine anyone having the time to do this

surveys/2021-annual-survey/questions.md
camelid: Just noting that there seems to be an extra "for" here.

nrc:
Addressed review comments

nrc:
I added a commit with one new question (IMO, a better version of the "What do you feel are the biggest challenges or problems for the Rust project? What could we do to improve adoption?" question) suggested by Nell, and some notes on similar questions we've mentioned in the design notes or previous years

PR #66, opened 2021-10-04
Add question about *when* the respondent learned Rust

rylev:
Let me know if the justification makes sense.

### DIFF
--- a/surveys/2021-annual-survey/questions.md
+++ b/surveys/2021-annual-survey/questions.md
@@ -123,9 +123,33 @@ Select all that apply:
 
 ## Your Rust experience
 
+### When did you learn to program in Rust?
+
+**Note**: while you may continue to try to improve your Rust skills, for this question assume "learning to program in Rust" means spending the majority of your time with Rust consuming learning materials or coding *in order to learn* (as opposed to achieving some other goal).
+
+Select one:
+
+- I'm still *actively* trying to learn Rust
+- Within the last year
+- Between 1 to 3 years ago
+- Between 3 to 6 years ago
+- More than 6 years ago
+
+> **justification**
+>
+> Useful for cohort analysis, i.e., for other questions we can understand if *when* someone learned Rust impacts their views on things.
+>
+> The time periods used as answers try to reflect the major "epochs" of Rust history (i.e., pre-1.0, 2015 edition, 2018 edition, etc.).
+
 ### How would you rate your Rust expertise?
 
-Select 0-4 (0 = I can't write or read Rust, 1 = I can write simple exercises in Rust, 2 = I can write useful, production-ready code but it is a struggle, 3 = I am productive writing Rust, 4 = I'm an expert)
+Select one:
+
+- I can't write or read Rust
+- I can write simple programs in Rust
+- I can write useful, production-ready code but it is a struggle
+- I am productive writing Rust
+- I'm an expert
 
 > **justification**
 >


surveys/2021-annual-survey/questions.md
```diff
@@ -123,9 +123,33 @@ Select all that apply:
...
+- Within the last year
+- Between 1 to 3 years ago
+- Between 3 to 6 years ago
+- More than 6 years ago
+
+> **justification**
+>
+> Useful for cohort analysis, i.e., for other questions we can understand if *when* someone learned Rust impacts their views on things.
+>
+> The time periods used as answers try to reflect the major "epochs" of Rust history (i.e., pre-1.0, 2015 edition, 2018 edition, etc.).
```
nrc, line 142: It might be easier to phrase it like this? I.e., what was the first version of Rust you used? Pre-1.0, 2015, 2018, currently learning, or something like that. Or maybe keep the years and ask "when did you start learning Rust"

nrc: I think it is an improvement, but I would still prefer to ask when people started learning Rust, since that is a point in time, rather than a period and so makes it less confusing to match with answers (and also removes the potential ambiguity around when a person is done learning)

rylev: I get the concern. Ideally we would want to know what period of time was most formative for their learning as it's not really important when they started or ended learning in their eyes but what was Rust like when they were deepest in "learning mode". This is obviously not really possible to ask, so I'm trying to ask it in a way that feels closest to that.

No matter which way we ask there will always be people who are unsure if which category they fall into.

surveys/2021-annual-survey/questions.md
```diff
@@ -123,9 +123,35 @@ Select all that apply:
...
+
+**Note**: while you may continue to try to improve your Rust skills, for this question assume "learning to program in Rust"
+means spending the majority of your time with Rust consuming learning materials or coding *in order to learn* (as opposed to achieving
+some other goal). If your learning process spans several of the listed time periods, pick the one where you felt you learned *the most*.
+
+Select one:
+
+- I'm still *actively* trying to learn Rust
+- Within the last year
+- Sometime between 2019 and last last year
```
nrc, line 136:
```suggestion
- Sometime between 2019 and last year
```

surveys/2021-annual-survey/questions.md
```diff
@@ -123,9 +123,35 @@ Select all that apply:
...
+### When did you learn to program in Rust?
+
+**Note**: while you may continue to try to improve your Rust skills, for this question assume "learning to program in Rust"
+means spending the majority of your time with Rust consuming learning materials or coding *in order to learn* (as opposed to achieving
+some other goal). If your learning process spans several of the listed time periods, pick the one where you felt you learned *the most*.
+
+Select one:
+
+- I'm still *actively* trying to learn Rust
+- Within the last year
```
nrc, line 135:
```suggestion
- During 2021```

File: surveys/2021-annual-survey/questions.md [hunk 4]
```diff
@@ -123,9 +123,35 @@ Select all that apply:
...
+- Sometime between 2019 and last last year
+- Sometime between 2015 and 2019
```
nrc, line 137: During 2016. 2017, or 2018

surveys/2021-annual-survey/questions.md
```diff
@@ -123,9 +123,35 @@ Select all that apply:
...
+- Before 2015
```
nrc, line 138: During 2015 or before

surveys/2021-annual-survey/questions.md
```diff
@@ -123,9 +123,37 @@ Select all that apply:
...
+**Note**: while you may continue to try to improve your Rust skills, for this question assume "learning to program in Rust"
+means spending the *majority* of your time with Rust consuming learning materials or coding *in order to learn* (as opposed to achieving
+some other goal). If your learning process spans several of the listed time periods, pick the one where you felt you learned *the most*.
+
+Select one:
+
+- I'm still *actively* trying to learn Rust
+- During 2021
+- During 2019 or 2020
+- During 2015, 2016, 2017, or 2018
```
rylev, line 137: The error messages [changed](https://blog.rust-lang.org/2016/09/29/Rust-1.12.html) quite a bit in late 2016 and thus learning Rust started looking much more similar to what it looks like today. We may want to distinguish between 2015/2016 and 2017/2018.

PR #67, opened 2021-10-04
Add a question about why to use Rust at work

rylev:
We also discussed adding a similar question for people who use Rust in their free time.

### DIFF
--- a/surveys/2021-annual-survey/questions.md
+++ b/surveys/2021-annual-survey/questions.md
@@ -649,6 +649,33 @@ Select one:
 > This question assess hiring sentiment. Although there is intrinsic uncertainty, it is easy to answer and forward looking.
 > It will also be interesting to see what the demand for Rust skills from companies is over time.
 
+### Rate how much the follow statements are reasons your team use Rust at work.
+
+Statements:
+
+- We use Rust for its performance characteristics.
+- We use Rust because we need to run software in a constrained or specialized environment.
+- Rust security and safety properties are important to us.
+- Rust allows us to build correct and bug free software.
+- We find it enjoyable or fun to program in Rust.
+- We already know Rust so it's our default choice.
+- We find it easy to prototype with.
+
+Rating:
+
+- Strongly agree
+- Agree
+- Neither agree nor disagree
+- Disagree
+- Strongly disagree
+
+> **justification**
+>
+> The Rust community and potential adopters of Rust have a lot of assumptions of why one would choose Rust for a project.
+> This question can help confirm or challenge our assumptions and see how they change over time.
+>
+> TODO: this question should only be asked for those who say they use Rust at work.
+
 ## Rust in education
 
 > This section is largely in service of the [contexts](./design/contexts.md) design document.


surveys/2021-annual-survey/questions.md
```diff
@@ -649,6 +649,33 @@ Select one:
 > This question assess hiring sentiment. Although there is intrinsic uncertainty, it is easy to answer and forward looking.
 > It will also be interesting to see what the demand for Rust skills from companies is over time.
 
+### Rate how much the follow statements are reasons your team use Rust at work.
+
+Statements:
+
+- We use Rust for its performance characteristics.
+- We use Rust because we need to run software in a constrained or specialized environment.
```
nrc, line 657: This feels a bit underspecified - do you mean resource-constrained? "Specialised" is pretty generic too

surveys/2021-annual-survey/questions.md
```diff
@@ -649,6 +649,33 @@ Select one:
...
+- Rust security and safety properties are important to us.
+- Rust allows us to build correct and bug free software.
```
nrc, line 659: I feel like there should be a "relatively" or "more" or something in here

surveys/2021-annual-survey/questions.md
```diff
@@ -649,6 +649,33 @@ Select one:
...
+- We find it enjoyable or fun to program in Rust.
+- We already know Rust so it's our default choice.
+- We find it easy to prototype with.
+
```
nrc, line 663: I think there should be an answer about expressivity of data types and the type system. Probably also a free form box for other major reasons

nrc (approved):

surveys/2021-annual-survey/questions.md
rylev: Resource constrained is what I had in mind, but I don't want to be too specific. I'm trying to get at the sentiment of "We need specific and direct control over how the end binary behaves, and Rust gives us that control". This differs from performance in that it's not about making it faster, consume less memory, etc., it's about making sure you have predictable and controllable behavior.

surveys/2021-annual-survey/questions.md
rylev: Can you explain how you were differentiate this from the "relatively correct and bug free aspect". Typically expressivity is preferred for better modeling the domain and thus leading to more correct programs. Some might also find this enjoyable in which case this is covered by one of the other possible answers.

surveys/2021-annual-survey/questions.md
nrc: Hmm, I feel that "predictable and controllable" is orthogonal to the environment in which the program is run. Whether you consider that separate from performance is an interesting question - IMO they are both about performance, one is minimizing resource usage and one is having predictable resource usage. But they are distinct enough to have separate answers. I wonder if we also need an answer specific to the environment, e.g., things like setting memory-mapped registers is impossible in a lot of languages and isn't about performance or resource usage.

PR #74, opened 2021-10-06
Add a question about activities that lead respondents to feel unwelcome

rylev:
This is a bit tricky of a question and something we should engage with the wider community and moderation teams on.

### DIFF
--- a/surveys/2021-annual-survey/questions.md
+++ b/surveys/2021-annual-survey/questions.md
@@ -481,6 +481,26 @@ Select all that apply:
 > particular we can use this information to do cohort analysis on highly "active"
 > community members in comparison to less active community members.
 
+### How often have felt *un*welcome in the Rust community?
+
+Activities:
+
+- *Official* Rust community forums or chats (e.g., users.rust-lang.org, the official Rust Discord, the Rust Zulip, etc.)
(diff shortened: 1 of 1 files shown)


surveys/2021-annual-survey/questions.md
```diff
@@ -481,6 +481,26 @@ Select all that apply:
 > particular we can use this information to do cohort analysis on highly "active"
 > community members in comparison to less active community members.
 
+### How often have felt *un*welcome in the Rust community?
```
nrc, line 484:
```suggestion
### How often have you felt *un*welcome in the Rust community?
```

surveys/2021-annual-survey/questions.md
```diff
@@ -481,6 +481,26 @@ Select all that apply:
...
+
+Activities:
+
+- *Official* Rust community forums or chats (e.g., users.rust-lang.org, the official Rust Discord, the Rust Zulip, etc.)
```
nrc, line 488: We should explicitly list some forums (esp r/rust) which are not official, since that is not always clear to people (or add that as a separate activity)

nrc (approved):
Approved, but I too would like to hear from others with more experience in this area.

PR #79, opened 2021-10-08
Add question about truisms

rylev:
With the removal of the questions in #78 this makes a question that attempts to capture the information we would otherwise be missing out on. Loosely these are all "truisms" about Rust that we might want to consider figuring out how much users of Rust agree with.

### DIFF
--- a/surveys/2021-annual-survey/questions.md
+++ b/surveys/2021-annual-survey/questions.md
@@ -204,6 +204,27 @@ Select all that apply:
 >
 > This question has led to surprising answers in the past that could help the lang team better prioritize features that make interop easier.
 
+### How much do you agree with the following statements
+
+Statements:
+
+- Rust provides a real benefit over other programming languages
+- Rust is complicated to use
+- Rust is risky to use in production
+- Rust makes me less productive
+
+Answers:
+
+- Strongly agree
+- Agree
+- Neither agree nor disagree
+- Disagree
+- Strongly disagree
+
+> **justification**
+>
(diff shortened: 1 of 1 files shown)


surveys/2021-annual-survey/questions.md
```diff
@@ -204,6 +204,27 @@ Select all that apply:
...
 > This question has led to surprising answers in the past that could help the lang team better prioritize features that make interop easier.
 
+### How much do you agree with the following statements
+
+Statements:
+
+- Rust provides a real benefit over other programming languages
+- Rust is complicated to use
+- Rust is risky to use in production
+- Rust makes me less productive
```
nrc, line 214:
```suggestion
- Rust makes me more productive
```

I think phrasing this positively is more consistent since I would expect that to be true.

surveys/2021-annual-survey/questions.md
```diff
@@ -204,6 +204,27 @@ Select all that apply:
 >
 > This question has led to surprising answers in the past that could help the lang team better prioritize features that make interop easier.
 
+### How much do you agree with the following statements
+
+Statements:
+
+- Rust provides a real benefit over other programming languages
+- Rust is complicated to use
```
nrc, line 212: Maybe also have 'Rust is difficult to learn'?

nrc (approved):

surveys/2021-annual-survey/questions.md
rylev: Yea we'll have to watch out for survivorship bias on all of these though.

PR #85, opened 2021-10-11
Add a question asking users to report on their skills in certain areas of low level software

### LINKED ISSUE #69: A question asking about computer science background
One piece of information that could be interesting for cohort analysis is whether the respondent feels comfortable with computer science topics. This can help us differentiate those who lean towards only have "practical" programming experience vs those who also have a theoretical base. Considering Rust's reputation as a more advanced or difficult language this could be illuminating. 

Some challenges include:
* Gauging the degree of knowledge could be difficult
* Those without formal education in CS can still be highly familiar with CS topics 
* Does this data *actually* help us understand the community better?

rylev:
Fixes #69 

Not sure I actually like this question. We might just end up making people feel bad if they report low comfort levels on all of these, and we may not even get actionable information out of it. Thoughts?

### DIFF
--- a/surveys/2021-annual-survey/questions.md
+++ b/surveys/2021-annual-survey/questions.md
@@ -974,6 +974,30 @@ Select one:
 - < 10 year
 - > 10 years
 
+### How would you rate your comfort with and understanding of the following concepts?
+
+Concepts:
+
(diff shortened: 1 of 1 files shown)


rylev:
I'd like to discuss this more before merging. I'm not convinced this is a good question.

PR #108, opened 2021-10-14
Add a commonly expressed worry for Rust

joshtriplett:


surveys/2021-annual-survey/questions.md
```diff
@@ -713,6 +713,7 @@ Select one:
 - Doesn't add a feature I want
 - Stagnation
 - Instability of the language
+- Development of the language not scalable/sustainable
```
rylev, line 716: Is this trying to capture the worry that the Rust project itself implodes or fails to keep up with the demands? I like this, but I'm not sure about the wording though I can't come up with a good alternative.

nrc, line 716:
```suggestion
- Development of the language is not scalable
```

I think having an answer here specifically about scaling is good. I think sustainability basically covers all of the answers, so this answer is better off without it.

joshtriplett, line 716:
```suggestion
- Development of the language not scalable
- Developers/maintainers of the language not paid/supported
```

nrc: lgtm

PR #115, opened 2021-10-15
Make the CPUs question a bit less technical

### LINKED ISSUE #87: On the primary machine you compile Rust code on, how many logical CPU threads do you have?
[Link to Question](https://github.com/rust-lang/surveys/blob/5685410a8e7c134015b2b111d069ac771508bfb3/surveys/2021-annual-survey/questions.md#on-the-primary-machine-you-compile-rust-code-on-how-many-logical-cpu-threads-do-you-have).

Unlike other questions this jumps into technical details immediately.
These things might not be familiar with everyone, particularly those from non-systems backgrounds.

Maybe this can be phrased in a more "normal" way first, before explaining what answer we're looking for.

E.g. "How powerful is the computer you usually work on?" followed by the explanation of getting info about logical CPUs

rylev:
Fixes #87

### DIFF
--- a/surveys/2021-annual-survey/questions.md
+++ b/surveys/2021-annual-survey/questions.md
@@ -20,7 +20,7 @@ We estimate it will take about 10-15 minutes to complete.
 
 ### Do you use Rust?
 
-Select one:
+Type: select one
 
 - Yes, I use Rust [`NEXT`](#your-rust-experience)
 - No, I don't currently use Rust, but I have in the past [`NEXT`](#for-previous-rust-users)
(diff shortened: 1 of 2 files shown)


surveys/2021-annual-survey/questions.md
```diff
@@ -198,16 +198,16 @@ Select all that apply:
 > specific answer appears frequently, we can add it to future surveys to reduce
 > the amount of work needed to process responses.
 
-### On the primary machine you compile Rust code on, how many logical CPU threads do you have?
+### On the primary machine you compile Rust code on, how many CPU do you have?
```
nrc, line 201:
```suggestion
### On the primary machine you compile Rust code on, how many CPUs are there?
```
Even if you decide to keep the original wording, we should use the plural of CPUs here.

PR #188, opened 2022-01-27
RFC: monthly micro surveys

nrc:
I would like to run monthly micro surveys, starting once we've had a chance to process the annual survey data. This PR is an unofficial RFC to determine if its a good idea. The PR adds a README describing monthly micro surveys, I expect we will iterate on that and can discuss in this PR.

cc @rust-lang/core since I believe new surveys are core team's remit, and for this to work we would need to use project resource (blog, Twitter account, etc) to promote the surveys.

### DIFF
--- a/surveys/monthly-micro/readme.md
+++ b/surveys/monthly-micro/readme.md
@@ -1,0 +1,65 @@
+# Monthly micro surveys
+
+## What is a micro survey?
+
+A small survey which is very low cost (in both time and cognitive load) to complete. They should be focussed on one topic (possibly with a first question to get context on the respondent for cohort analysis), should take fewer than five minutes to complete, be three or fewer questions (all multiple choice rather than free-form, though we might ask a single free-form question instead), not require too much thinking, research, or permission (from employers, etc.), and be fun to answer. See below for some examples.
+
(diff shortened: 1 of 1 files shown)


apiraino:
I like this proposal a lot. As we are designing the 2023 survey I see now how it is difficult to compress data (questions + options for answers) in it and make it an all-encompassing tool to learn about _everything_ (which is unrealistic).

Minus the "monthly format", I think everything in this proposal is within our reach. I see these mini surveys more at an irregular cadence, happening when we have _really_ something to ask, otherwise we might end up _looking for a topic_ to respect a self-imposed monthly cadence. We could poll the community about the topics to be surveyed upon :sweat_smile: 

Also, this [Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/Rust.20Survey/near/396174184) for more thoughts about how a more focused survey coold look like.

I'd like to circle back on this after the annnual survey bandwagon is gone.

apiraino:
Another interesting question [I noticed on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/301329-t-devtools/topic/Having.20a.20CI.20subteam.3F/near/412221832): "Do people want an official CI Rust Github Action?"

Kobzol:
Closing in favour of https://github.com/rust-lang/surveys/pull/257.

PR #189, opened 2022-02-08
Policy proposal: data sharing

nrc:
This document describes how data from the annual survey should be shared.

r? @rust-lang/core cc @rylev


documents/data-sharing.md
```diff
@@ -0,0 +1,58 @@
...
+
+This document describes how data from the annual survey should be shared.
+
+## Objectives
+
+In priority order:
+
+* Protect the privacy of respondents.
+* Provide high quality data (unbiased data which is unlikely to be misinterpreted, c.f., raw data).
+* Provide useful and actionable data to the Rust project (Core team, Foundation, teams and working groups, contributors).
```
Mark-Simulacrum, line 11: I don't think it's worth singling out the core team or Foundation here -- it has as much use for the data as other Rust teams. If there is something to emphasize, it might be a prioritization of the team/working group/project group access to data over individuals (e.g., I might see us asking each team to submit a list of questions, but that the intent is that those questions are related to their work as a team).

documents/data-sharing.md
```diff
@@ -0,0 +1,58 @@
...
+* Provide useful data for users and potential users of Rust and providers of Rust tools and services.
```
Mark-Simulacrum, line 12: I might try to reformulate this as seeking to provide summaries to this group -- I am not sure it should be a goal to share un-digested 'raw' data (even if it's in e.g. "this question had X% responding with Y column" aggregate, particularly with fully external parties to the project.

Mostly this is to avoid a situation where we accidentally over-share, so maybe this is just 'low enough' that priority 1 (privacy) is sufficient, but I might still reformulate. I'm not sure we should be spending time as a project on assembling data for fully public consumption.

documents/data-sharing.md
```diff
@@ -0,0 +1,58 @@
...
+* Provide data in a timely manner.
+* Provide marketing for Rust.
+* Avoid overloading people working directly with the survey data (note that there were over 9000 completed surveys and multiple opportunities for free form responses).
+
+Non-goals:
+
+* We will not aim to make money by selling the data in any form.
+* We will aim to not privilege companies which employ people with access to survey data (raw or processed). However, we acknowledge that controlling widely shared data is impossible.
```
Mark-Simulacrum, line 20: I might s/We/The project/ or similar just to make clear who is providing this (non)goal.

documents/data-sharing.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+## Blog post
+
+The survey working group will author a blog post for publication on the main Rust blog as soon as possible after the survey is closed. The blog post should acknowledge respondents, and summarise highlights of the survey data. Highlights should be chosen based on being useful or interesting to a general audience (i.e., Rust users and non-users). The blog post should not aim to be a complete summary of the data.
```
Mark-Simulacrum, line 25: I might add something to the effect of "The primary intent of the blog post is to communicate interest and usage of Rust, including drilling into interesting categories/use-cases" - maybe there's some other primary goal, but I think something along these lines might be helpful to try and narrow in on in terms of our plan for the blog post.

Mark-Simulacrum (commented):
Wanted to leave some initial feedback, but didn't go through the full doc yet.

documents/data-sharing.md
nrc: Hmm, I was worried that not writing Foundation explicitly might exclude it since it is not technically a team or working group. Core team I guess is duplication. To be clear this list is just spelling out who exactly is included in 'the Rust project', it's not meant to be in priority order.

documents/data-sharing.md
nrc: I am trying to list the objectives of sharing in a general sense, so here I mean 'data' in an abstract way, rather than specifically the survey data.

> Mostly this is to avoid a situation where we accidentally over-share, so maybe this is just 'low enough' that priority 1 (privacy) is sufficient, but I might still reformulate. 

Yeah, these objectives are in priority order, so we shouldn't try to achieve any of the others at the expense of privacy.

> I'm not sure we should be spending time as a project on assembling data for fully public consumption.

I disagree with this though - if it benefits the project, by helping people adopt Rust or helps persuade companies to produce tools for Rust, etc., then that seems like a good use of our time.

nrc:
@Mark-Simulacrum thanks for the comments, I've made changes to the text.

documents/data-sharing.md
```diff
@@ -0,0 +1,58 @@
...
+
+In priority order:
+
+* Protect the privacy of respondents.
+* Provide high quality data (unbiased data which is unlikely to be misinterpreted, c.f., raw data).
+* Provide useful and actionable data to the Rust project (i.e., teams and working groups, contributors, and the Foundation).
+* Provide useful data for users and potential users of Rust and providers of Rust tools and services.
+* Provide data in a timely manner.
+* Provide marketing for Rust.
+* Avoid overloading people working directly with the survey data (note that there were over 9000 completed surveys and multiple opportunities for free form responses).
```
rylev, line 15: Nit: this refers to the 2021 annual community survey (i.e., 9000 completed surveys) but doesn't explicitly name it - this may be confusing in the future.

documents/data-sharing.md
```diff
@@ -0,0 +1,58 @@
...
+
+Non-goals:
+
```
rylev, line 18: I believe an additional goal will not be to assist commercial products - e.g., if a company sells Rust based tooling, it is not a goal for the survey to explicitly support those projects with data.

documents/data-sharing.md
```diff
@@ -0,0 +1,58 @@
+# Annual survey data sharing
```
rylev, line 1: Can we refer to it as the "annual community survey" (also in the name of the file). We are likely to do many more types of surveys and I believe we want to limit the scope to just the annual community survey. We may also additionally want to link to the 2021 survey as an example.

documents/data-sharing.md
```diff
@@ -0,0 +1,58 @@
...
+
+* Protect the privacy of respondents.
+* Provide high quality data (unbiased data which is unlikely to be misinterpreted, c.f., raw data).
+* Provide useful and actionable data to the Rust project (i.e., teams and working groups, contributors, and the Foundation).
+* Provide useful data for users and potential users of Rust and providers of Rust tools and services.
+* Provide data in a timely manner.
+* Provide marketing for Rust.
+* Avoid overloading people working directly with the survey data (note that there were over 9000 completed surveys and multiple opportunities for free form responses).
+
+Non-goals:
```
rylev, line 17: I'm unsure how to word this, but I also believe we should be explicit that the survey is a best attempt and not meant to be a rigorous statistical study. This means we may ask questions that to the trained survey creator and statistician are not necessarily ideal.

documents/data-sharing.md
```diff
@@ -0,0 +1,58 @@
...
+
+This initial report will be sent to all members of Rust teams, working groups, and the Rust Foundation. Recipients will be asked not to share the report outside the Rust project.
+
+The survey working group will present the report to the Rust project and answer questions, for example in the Cross-Team Collaboration Fun Times meeting (note that this meeting is public and thus care must be taken not to discuss and sensitive information).
+
+A reduced version of the report with any sensitive information removed will be produced after the initial report is distributed. With core team approval, this report will be shared publicly, but not strongly advertised (i.e., a post on the Inside Rust blog, but not the main blog).
+
+Unresolved questions: is the core team the right group to approve the public report? Should we specify what we consider sensitive?
```
rylev, line 38: I think diving a bit into what might be considered sensitive would be helpful.

rylev (commented):
This seems like a good start. I'd like to give it another read soon before approving.

documents/data-sharing.md
nrc: I think there is a bit of a fine line here. On the one hand, I don't think we should accept requests for custom stats, etc. from companies, and I don't think we should be producing statistics in general with the goal of helping companies sell more products. But, I think producing statistics that might be helpful for a person or team to persuade their management to resource tooling for Rust (commercial or not) seems like something which we should strive to do since it has obvious benefit for the community as a whole

nrc:
Thanks Ryan, comments addressed!

documents/data-sharing.md
rylev: The key is broad applicability. We can take suggestions of what we should share from the community (including commercial products), but we should evaluate those suggestions on how applicable they are to the wider community. 

I believe we're on the same page here, so I think it's just that I believe this should be more explicitly laid out.

documents/data-sharing-annual-community-survey.md
```diff
@@ -0,0 +1,70 @@
...
+* Provide data which is generally useful for the community (users and potential users of Rust, and providers of Rust tools and services). Note that this does not mean producing data to be useful for specific segments of the community.
+* Provide marketing for Rust.
+* Avoid overloading people working directly with the survey data (note that in 2021 there were over 9000 completed surveys and multiple opportunities for free form responses).
+
+Non-goals:
+
+* Making money by selling the data in any form.
+* The project will aim to not privilege companies which employ people with access to survey data (raw or processed). However, we acknowledge that controlling widely shared data is impossible.
+* Directly supporting commercial products or companies with data.
+* Produce a rigorous or complete statistical study of survey results.
```
rylev, line 22: I still think we're missing a non-goal like:
* provide individuals or specific groups with data that is likely only beneficial to them and not to the wider community 

We don't want to be in the business of servicing ad-hoc requests with the exception of ad-hoc requests for project teams (and even then we'll want to set limits).

documents/data-sharing-annual-community-survey.md
```diff
@@ -0,0 +1,87 @@
...
+
+Unresolved question: should there be a time limit?
+
+## Raw data
+
+The full data set of survey responses will never be widely shared. It should be accessible only to the subset of the survey working group responsible for the annual community survey, contractors engaged to help process and analyse the data, and translators (who should only have access to the data they are translating, not the whole data set). These groups may be expanded in the future, but should be kept to a minimum. Everyone with access should have completed a signed confidentiality agreement with the Rust Foundation.
```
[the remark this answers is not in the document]
rylev: The more we give access to raw data, the more we expose ourselves to the risk of individuals making honest mistakes and leaking data they shouldn't. I understand the desire to want to get out of folks way, but the "summary data" is still in a form that allows arbitrary querying. 

The only thing that the raw data allows for that summary data does not is analysis of free form text. While I can imagine insight coming from analyzing free from text without having a known goal ahead of time, this is a *lot* of work, and I don't imagine many individuals wanting to do this. If one is really that interested in the data, they should probably be on the survey team themselves and sharing their analysis with others.

nikomatsakis:
Does summary data allow for correlating data across questions?

documents/data-sharing-annual-community-survey.md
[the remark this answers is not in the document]
nikomatsakis: I agree that people are unlikely to want to "deep dive" into freeform data. That said, I've personally also found it useful to just read all the freeform responses from time to time, or at least big chunks of them, to get a sense for the "kinds of things" that people say.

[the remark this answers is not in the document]
rylev: > Does summary data allow for correlating data across questions?

No, it doesn't, but that's a limitation of our current tooling not something that is limited by this policy. I think it makes more sense for us to fix the tooling to allow for this rather than opening up the data.

> to get a sense for the "kinds of things" that people say.

This is something the survey team would provide. The question becomes: is the risk associated with opening up data to others worth allowing others outside the survey team the ability to participate in this process in parallel?

documents/data-sharing-annual-community-survey.md
```diff
@@ -0,0 +1,87 @@
...
+
+In priority order:
+
+* Protect the privacy of respondents.
+* Provide high quality data (unbiased data which is unlikely to be misinterpreted, c.f., raw data).
+* Provide useful and actionable data to the Rust Project (i.e., teams and working groups, contributors) and Foundation.
+* Provide data in a timely manner.
+* Provide data which is generally useful for the community (users and potential users of Rust, and providers of Rust tools and services). Note that this does not mean producing data to be useful for specific segments of the community.
+* Provide marketing for Rust.
+* Avoid overloading people working directly with the survey data (note that in 2021 there were over 9000 completed surveys and multiple opportunities for free form responses).
```
pnkfelix, line 15: Nitpick: I'm glad "avoid overloading people working directly with the survey data" is listed here. I suspect it is not prioritized highly enough.

E.g. I think it is higher priority than providing marketing for Rust. It may even be higher priority than "Provide data that is generally useful to the community."

documents/data-sharing-annual-community-survey.md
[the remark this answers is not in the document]
rylev: Perhaps "arbitrary querying" would not be allowed, but I can definitely imagine a system that allows correlating summary data as long as we limit queries to those that don't limit the data set beyond a certain point (i.e., don't allow queries that produce sets of less than ~500). 

Either way, the question at hand is whether we allow all Rust team members access to the data (provided a signed NDA) or not. I'd lean towards keeping raw data in the smallest set of hands possible, since the worst case scenario of leaking data is pretty bad (i.e., once data is leaked, it's leaked for good, and the only recourse we have is to enforce an NDA against a team member which sounds horrible). 

If we find that this set up is very painful, we can change the policy and go through the proper due diligence of informing survey respondents of this.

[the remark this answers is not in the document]
nikomatsakis: I'm trying to decide how hard to push here. My thinking is like this: if we are going to seriously try to get use of the survey data (and I'm not sure whether it's the right tool for that, to be quite honest), I at least think the way I would do that is to sit down with the questions and read through them. I would find interesting questions that I'd like to understand better, and start formulating queries and formulations against the data to do that (e.g., people who said they are experts come from these domains, etc).

It's likely I would not want access to all the data. I would probably want access to "summary data" from free form comments ("of the known keywords, did this person reference it?"). I could imagine caring what country someone is from, but that seems fairly unlikely. 

In my *ideal* world, I would be doing these queries via some sort of web interface to a cool data analysis platform, so I would never actually have the data on my local machine. Thus the question of exporting would be pretty moot. 

In any case, that workflow could certainly work with survey team members as intermediaries, but it seems like a waste of everyone's time. And the problem is that it is hard for me to formulate queries up front, the choice of what to ask is maybe driven by the answers to previous questions.

Certainly I am influenced by this feeling that this data is -- by and large -- fairly impersonal, and so while I agree that leaking the whole data set would be bad, I feel like people's expectations of privacy are probably fairly low. I could be wrong in this assessment. 

All of that said: I would like to do the above for this year's data, I'm aware I didn't do that =)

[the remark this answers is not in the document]
nrc: So, I think that there is a need for people to do that kind of deep dive into the data in general and that it would be beneficial to do. However, I think there is a trade-off here against professional treatment of people's data and that means having guard rails at least for how data is accessed and in particular minimising the number of people who have access to the raw data (even though the data is fairly impersonal, I think that there is still an expectation of privacy. I think part of being a 'good citizen' is that the Rust project treats the collected data with the highest level of respect for privacy and we follow best practices, etc even where that seems excessive. To give some concrete examples of hypothetical scenarios: 1, a respondent leaves a comment like "nrc is a big fat doodihead" in a free form answer. Given they come from a country with only 6 respondents and their answers to contribution questions indicate they are a regular contributor, nrc can identify them and is then (consciously or unconsciously) biased against them in future work. 2, 50% of respondents who speak Portuguese said they feel unwelcome due to their sexual orientation, this is mentioned as a joke by a team member at a meetup in Brazil and is overheard by someone with connections to a homophobic hate group, Rust meetups become targets for harassment.)

I do think that *someone* should do the kind of work you're describing and thus there is a role for that in this policy. Currently the survey WG, though I could imagine that being a product team in the future if such a thing ever existed. I strongly believe that it would be a mistake to expand access to the raw data to a large group of people (e.g., all team members or even all team leads). It is not unreasonable that we might want to expand the group beyond the survey WG, but I think that group should be well-defined, minimal, and have good reason to want to work with the data in this manner (i.e., curiosity is not a good enough reason). I don't have any suggestions for defining such a group, but I'd be happy to hear some.

Alternatively, you could join the survey WG to do this kind of analysis. I think expanding the number of people in the WG to analyse the data should be expanded, and that membership for the purpose of analysis without obligation to writing next year's survey, etc. is reasonable.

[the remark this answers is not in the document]
nikomatsakis: I'm willing to let it stand as the survey wg (as it happens, I am currently a member anyway, but that's neither here nor there) and see how much trouble it is. My concern is that the result will be lower engagement with the data overall, but there are other ways to work on that. 

I still think the danger is minimal, particularly with some effort at education, and that team leads in particular (though also members) have clear good reasons to want to work with the data in this fashion (they are trying to decide what work to prioritize and how to direct their energies). That said, I'm  *not* sure how useful the data is in practice for that purpose, and in particular the correlated data (versus the simple summary). So I'd be happy to start with the summary and see where we go.

rylev (approved):
This has been discussed enough for me to feel comfortable using this as the policy. We can revisit this policy as we gain more experience with it.

PR #212, opened 2022-10-11
2022 cuts

nrc:
r? @rylev as discussed (I think)

### DIFF
--- a/surveys/2022-annual-survey/questions.md
+++ b/surveys/2022-annual-survey/questions.md
@@ -26,40 +26,11 @@ Type: select one
 
 ## For previous Rust users
 
-### As you have indicated that you're no longer using Rust, what prompted you to participate in this survey?
-
-Type: select all that apply (optional)
-
-- I plan to return to using Rust in the future
-- I consider myself part of the Rust community
-- Specifically to provide feedback on why I stopped using Rust
-- To provide feedback on Rust in general
-- Curiosity
-- Other (open response)
-
-> **justification**
->
-> Useful in understanding why non-users contribute;
-
 > **SURVEY FLOW**
 >
 > Skip to `## Your opinions about Rust` section
 
 ## For non-Rust users
-
-### As you have indicated that you have not used Rust, what prompted you to participate in this survey?
-
-Type: select all that apply (optional)
-
-- I plan to use Rust in the future
-- I consider myself part of the Rust community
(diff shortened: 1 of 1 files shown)


surveys/2022-annual-survey/questions.md
```diff
@@ -26,41 +26,12 @@ Type: select one
...
->
-> Useful in understanding why non-users contribute;
-
 > **SURVEY FLOW**
 >
 > Skip to `## Your opinions about Rust` section
 
 ## For non-Rust users
 
-### As you have indicated that you have not used Rust, what prompted you to participate in this survey?
```
rylev, line 50: Perhaps we want to replace these with a TODO just so we're sure we don't forget that we should either make new questions here or delete these sections entirely.

surveys/2022-annual-survey/questions.md
```diff
@@ -669,12 +596,6 @@ Aspects:
 - Disk space usage (e.g., the size of target folder)
 - Bugs in the compiler (i.e., ICEs a.k.a. internal compiler errors, miscompilations, etc.)
 - Compiler error messages
-- IDE experience
-- Debugging experience
-- Available tools and support
-- Async programming
-- GUI development
-- Rust language and standard library documentation
```
rylev, line 677: Thinking more about it, I'm not sure about this...

rylev (changes requested):
I got cold feet on removing some of the "aspects of Rust" that we ask respondents to answer... Otherwise, looks great (minus a small nit).

surveys/2022-annual-survey/questions.md
nrc: Note to self (from Wesley): could we get respondents to rank these in priority order, to discover how people would feel about sacrificing, e.g., disk space for compile times?

surveys/2022-annual-survey/questions.md
nrc: I have the TODOs in a separate branch

surveys/2022-annual-survey/questions.md
rylev: I think having all of these rated would be useful.

rylev: @nrc can you add these back, and then we can discuss separately whether to cut them?

nrc:
Tools stuff restored

rylev (approved):

PR #213, opened 2022-10-17
Suggested changes for 2022 annual survey

nrc:
cc @rylev

### DIFF
--- a/surveys/2022-annual-survey/questions.md
+++ b/surveys/2022-annual-survey/questions.md
@@ -26,11 +26,51 @@ Type: select one
 
 ## For previous Rust users
 
+### Why did you stop using Rust?
+
+Type: select all that apply
+
+- Rust did not help me achieve my goals
+- Missing language features
+- Missing libraries
+- Missing tools
+- Too difficult to learn
+- Community was rude, unwelcoming, or otherwise off-putting
+- I prefer to use another language
+- I no longer have the opportunity to use Rust due to factors outside of my control
+- Other
+
+### Tell us more:
+
+Type: free text
+
+
 > **SURVEY FLOW**
 >
 > Skip to `## Your opinions about Rust` section
 
+TODO there are some questions before this which we could ask non-users, how do we make this work with the software?
+
 ## For non-Rust users
+
+### Why don't you use Rust?
+
+Type: select all that apply
+
+- Rust did not help me achieve my goals
+- Missing language features
+- Missing libraries
+- Missing tools
+- Too difficult to learn or learning will take too much time
+- Community was rude, unwelcoming, or otherwise off-putting
+- I prefer to use another language
+- Due to factors outside of my control
+- Haven't got around to it
+- Other
+
+### Tell us more:
+
+Type: free text
 
 > **SURVEY FLOW**
 >
@@ -168,7 +208,7 @@ Type: select all that apply (optional)
 - A specific version of nightly
 - Custom fork
 - I don't know
-- Other (open response)
+- Other
 
 > **justification**
 >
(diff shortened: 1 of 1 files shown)


surveys/2022-annual-survey/questions.md
```diff
@@ -26,40 +26,62 @@ Type: select one
...
+Type: select one
+
+- Yes
+- No
 
 > **SURVEY FLOW**
 >
 > Skip to `## Your opinions about Rust` section
 
+TODO there are some questions before this which we could ask non-users, how do we make this work with the software?
```
rylev, line 57: Which questions do you mean?

surveys/2022-annual-survey/questions.md
```diff
@@ -26,40 +26,62 @@ Type: select one
...
-Type: select all that apply (optional)
+Type: select all that apply
 
-- I plan to return to using Rust in the future
-- I consider myself part of the Rust community
-- Specifically to provide feedback on why I stopped using Rust
-- To provide feedback on Rust in general
-- Curiosity
-- Other (open response)
+- No benefit or not appropriate for my projects
```
rylev, line 33: I'm not sure if this category covers folks who simply don't have the time to use it for their projects. For example, someone who is working on web app front ends maybe *could* use Rust, but it's not really super noteworthy that they are not. I wonder if it's the fact that this choice makes it seem much more negative than it might be. 

Perhaps we could split this into the following:
* I no longer have the opportunity to use Rust due to external factors
* Rust did not help me achieve my goals

surveys/2022-annual-survey/questions.md
```diff
@@ -277,79 +285,35 @@ Type: select all that apply (optional)
...
-- I have participated in Rust community forums or chats (e.g., users.rust-lang.org, Rust Discord, a local Rust chat community, etc.)
-- I have attended a Rust meetup or conference (virtual or in-person)
-
-> **justification**
->
-> We'd like to get a picture of _how_ people participate in the Rust community. In
-> particular we can use this information to do cohort analysis on highly "active"
-> community members in comparison to less active community members.
-
 ### Roughly how often do you contribute to the Rust project?
```
rylev, line 288:
```suggestion
### Roughly how often do you contribute to the Rust community?
```

surveys/2022-annual-survey/questions.md
```diff
@@ -277,79 +285,35 @@ Type: select all that apply (optional)
...
-- Open an issue on any repo in the rust-lang GitHub organization
-- Contribute code changes (including tests) to the Rust compiler (rust-lang/rust)
-- Contribute code changes (including tests) to any other project in the rust-lang GitHub organization
+- Produce informational content about Rust (e.g., blogged, live streamed, made a YouTube video, presented at a conference/meetup, etc.)
+- Consume informational content about Rust (e.g., blogs, live streams, YouTube videos, etc.)
+- Read official Rust communication channels (e.g., This Week in Rust, the official Rust blog, the Rust Twitter account, etc.)
+- Participate in conversations about Rust on social media or websites (Hacker News, r/rust, Twitter, LinkedIn, etc.)
+- Attend a Rust meetup or conference (virtual or in-person)
+- Write, comment on, contribute to discussion of, or provide edits to an open RFC
+- Discuss the Rust project in an official chat (internals.rust-lang.org, Zulip, or Discord)
```
rylev, line 300:
```suggestion
- Discuss the Rust project in an official chat (internals.rust-lang.org, rust-lang Zulip, etc.)
```

surveys/2022-annual-survey/questions.md
```diff
@@ -364,60 +328,96 @@ Activities:
 
 Choices:
 
-- I've *often* felt welcome
-- I *sometimes* feel welcome
-- I *never* feel welcome
+- I feel welcome
+- I feel neutral
+- I feel unwelcome
```
rylev, line 333: I wonder if folks who *sometimes* feel unwelcome will respond differently than folks who *often* feel unwelcome. I am afraid we miss important signal by asking things this way...

surveys/2022-annual-survey/questions.md
```diff
@@ -364,60 +328,96 @@ Activities:
...
-- I've *often* felt unwelcome
-- I *sometimes* feel unwelcome
-- I *never* feel unwelcome
-- I've never participated in this activity
+### Are you personally using Rust at work?
+
+Type: select one
+
+- Yes, for the majority of my coding
+- Yes, it's one of a number of languages I use and I use it regularly
```
rylev, line 357: Is there a meaningful difference between these two?

surveys/2022-annual-survey/questions.md
```diff
@@ -364,60 +328,96 @@ Activities:
...
 > **justification**
 >
-> We'd like to know where people are feeling unwelcome and the degree to which they are feeling unwelcome. This can 
-> help us better understand the free from responses that will come in the next question.
+> We want to establish what percentage of those who could possibly use Rust in a professional setting
+> are using Rust in a professional setting. This is most interesting over time.
+> Answers to this question should be combined with whether the respondent has ever used Rust.
+
+
+### Which of the following statements are reasons why your team uses Rust at work
```
rylev, line 368: We ask about people "personally" using Rust at work, and then jump to their "team". Not sure if that's a problem though.

surveys/2022-annual-survey/questions.md
```diff
@@ -463,26 +463,29 @@ If you've previously answered that your company is not actively using Rust, you
 
 Type: select all that apply (optional)
 
+TODO check last year's responses
```
rylev, line 466: I'd personally prefer to not capture TODOs like this in the document itself.

surveys/2022-annual-survey/questions.md
```diff
@@ -645,21 +567,15 @@ Statements:
 - Rust makes me more productive
 - Rust is fun to use
 
-Rating:
```
rylev, line 648: Are we sure we're not losing important signal by allowing people to actively disagree?

surveys/2022-annual-survey/questions.md
nrc: Quite a lot of the survey, e.g., asking about Rust at work (just because they are a non-user doesn't mean their company doesn't use Rust)

surveys/2022-annual-survey/questions.md
nrc: IMO, it's a clearer signal because we don't let people cop out and choose the middle option. We might lose some signal here, but I think that simplifying for both the respondents and in the results is worth it

surveys/2022-annual-survey/questions.md
nrc: I'm not sure the difference means much in practice. I guess that if we were prioritising between venues it would, but I don't think we'd do that. Also the numbers are so small that for understanding the results it's easier to bucket these response together anyway

surveys/2022-annual-survey/questions.md
```diff
@@ -26,12 +26,65 @@ Type: select one
...
+Type: select all that apply
+
+- Rust did not help me achieve my goals
+- Missing language features
+- Missing libraries
+- Missing tools
+- Too difficult to learn
+- Community was rude, unwelcoming, or otherwise off-putting
+- I prefer to use another language
+- I no longer have the opportunity to use Rust due to external factors
```
rylev, line 40:
```suggestion
- I no longer have the opportunity to use Rust due to factors out of my control
```

Perhaps this reads a bit more clearly?

surveys/2022-annual-survey/questions.md
```diff
@@ -26,12 +26,65 @@ Type: select one
...
+- Other
+
+### Tell us more:
+
+Type: free text
+
+### Would you be likely to use Rust if the above issues were addressed
```
rylev, line 47: I'm not quite sure what we hope to learn from this? Are we suspecting that some users will answer the question in some way that we can ignore because they will also say that they won't use Rust even if that issue is ignored?

surveys/2022-annual-survey/questions.md
```diff
@@ -26,12 +26,65 @@ Type: select one
...
+Type: select all that apply
+
+- Rust did not help me achieve my goals
+- Missing language features
+- Missing libraries
+- Missing tools
+- Too difficult to learn or learning will take too much time
+- Community was rude, unwelcoming, or otherwise off-putting
+- I prefer to use another language
+- I no longer have the opportunity to use Rust due to external factors
```
rylev, line 73: I don't think this makes sense in this question - and definitely not how it's currently worded.

surveys/2022-annual-survey/questions.md
rylev: Perhaps just saying "why you and your team"?

surveys/2022-annual-survey/questions.md
```diff
@@ -298,60 +330,95 @@ Activities:
...
 ### To what extent is Rust currently being used by your company?
 
 Type: select one
 
-- My company uses Rust for a large portion of production projects
-- My company uses Rust for a small portion of production projects
-- My company uses Rust only for non-production projects (e.g., tooling)
-- My company has actively experimented with Rust
-- My company has seriously considered, but not experimented with, using Rust
+- My company uses Rust
```
rylev, line 420: It would be nice if we could still get some signal between "my company uses Rust but it's not an important vs. unimportant piece of our tech stake"

surveys/2022-annual-survey/questions.md
```diff
@@ -401,22 +468,25 @@ Type: select all that apply (optional)
 - Blockchain
 - Cloud computing applications
 - Cloud computing infrastructure or utilities
+- CLI
```
rylev, line 471: I would spell this out. I've talked with folks experienced in the industry who don't know the term CLI as "command line interface".

surveys/2022-annual-survey/questions.md
```diff
@@ -572,21 +568,15 @@ Statements:
...
-- Disagree
-
 > **justification**
 >
 > There are several "truisms" about Rust that we'd like to get perspective on how true these are for users of Rust.
 >
 > Note that answers here can be subject to survivorship bias and so extra care should be taken with interpreting results.
 
-### In your opinion, how do you find the following aspects of Rust?
+### In your opinion, how would you prioritise work on the following aspects of Rust?
```
rylev, line 577: I'm not sure I understand the question. Are expecting for the user to put these items in order? Does survey hero support this?

rylev (commented):
Getting closer, but I still have some questions

rylev (approved):

PR #219, opened 2022-11-02
Add a question on oldest Rust version used

nrc:
I think this is a useful question with a specific use case. There are 67 answers, so the format only makes sense if we can use a drop down box or other concise mechanism. If not, then we should refactor the answer (bucketed answers? Free text?).

r? @rylev


rylev (approved):
Looks good. Survey hero supports drop down menus so this shouldn't be an issue in practice.

PR #226, opened 2022-12-06
Add Hong Kong SAR to the list of places to live

zoie-bg:
The place where is live is not listed as an option.  From Wikipedia (https://en.wikipedia.org/wiki/Hong_Kong):

Hong Kong, officially the Hong Kong Special Administrative Region of the People's Republic of China (abbr. Hong Kong SAR)

### DIFF
--- a/surveys/2022-annual-survey/questions.md
+++ b/surveys/2022-annual-survey/questions.md
@@ -702,6 +702,7 @@ Type: select one (optional)
 - *all UN member states*
 - *two observer states (Vatican City and Palestine)*
 - Taiwan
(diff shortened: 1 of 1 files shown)


rylev:
Thanks for your feedback. In an effort to keep the list as manageable as possible, we have been relying on folks to use the "other" option if they do not find an option that suits them. "Other" answers will be counted in the same way as answers in the drop-down list.

For next year's survey, we'll make sure to revisit this to see how we can better balance the ease with which we collect data with the needs of respondents to clearly and correctly express where they live.

PR #241, opened 2023-10-09
Add question about build systems

### LINKED ISSUE #237: Add question about cargo usage
- Do you use Cargo?
- Do you use some other build system?
- Do you combine Cargo and some other build system?
  - Which system drives the compilation?

apiraino:
closes #237 

Open to improve wording, question placing and markdown fixes (for the SurveyTeam parser)

### DIFF
--- a/surveys/2023-annual-survey/questions.md
+++ b/surveys/2023-annual-survey/questions.md
@@ -594,6 +594,21 @@ Aspects:
 >
 > This question gathers data on the communities perceptions of certain aspects of Rust at this point in time.
 
+### How do you build your Rust projects?
+
+Type: select one (optional)
+
+Aspects:
+
+- I use Cargo
+- I use some other build system
+- I combine Cargo and another build system
+- If you combine Cargo with (or just use) other build systems, which ones? (open response)
+
+> **justification**
+>
+> cargo team expressed interest in this, see https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/Rust.20survery.202023/near/393816653
+
 ### Do you agree with the following statements on Rust stability?
 
 Type: matrix (optional)


surveys/2023-annual-survey/questions.md
```diff
@@ -594,6 +594,21 @@ Aspects:
 >
 > This question gathers data on the communities perceptions of certain aspects of Rust at this point in time.
 
+### How do you build your Rust projects?
+
+Type: select one (optional)
```
Kobzol, line 599: I think that this might be a multiple select question, people can have multiple projects, use Cargo on some of them, use a different build system elsewhere.

surveys/2023-annual-survey/questions.md
```diff
@@ -594,6 +594,21 @@ Aspects:
...
+
+Aspects:
+
+- I use Cargo
+- I use some other build system
+- I combine Cargo and another build system
+- If you combine Cargo with (or just use) other build systems, which ones? (open response)
```
Kobzol, line 606: One question mentioned by @joshtriplett is who drives the build system. But I'm not sure if we can express it properly. Already this last answer is quite complicated to understand. My attempt:

```
- Cargo alone
- Cargo drives another build system
- Another build system drives Cargo
- Another build system without Cargo
- What build system (apart from Cargo) do you use? (open)
```

Kobzol (commented):
I would maybe put this around/after `### What is the oldest version of Rust you use for any development task?`, before the IDE question, before this talks about how people build/develop their projects.

surveys/2023-annual-survey/questions.md
apiraino, line 606: Yes, the last option is not super clear. But I don't have the expertise to suggest examples. f.e. I don't have the context to appreciate the difference between your option 3 and 4.

```suggestion
- I just use Cargo
- I use some another build system (please specify) (open response)
- I combine Cargo and another build system (please specify) (open response)
```

:thinking: would it make it a bit clearer?

Kobzol: If we can have two open responses, then this seems nice! :)

apiraino: > If we can have two open responses, then this seems nice! :)

It seems we cannot (tested on a SurveyHero account test account)

apiraino, line 606:
```suggestion
- I just use Cargo
- I just use some another build system
- Please specify which build system are you using instead of (or with) Cargo (open response)
- ```

Having only one open option, I cannot make it clearer than this :)

apiraino: Having only one open option, I cannot make it clearer than this :)

Kobzol: Since the question now allows selecting multiple answers, I think that `just` doesn't make sense here. If you use just Cargo, you'll click only on Cargo. If you use both Cargo and another build system, you check both.

There's a third option, and that's combining Cargo and another build system, which I would keep as an answer (`I combine Cargo and another build system`).

apiraino: Ok, I think I'm basically back to square one.

Kobzol: Tbh I think that changing it to `all that apply` was all that was needed, in my opinion :sweat_smile:

PR #263, opened 2024-06-18
Add Leadership Council annual survey

ehuss:
This adds a micro-survey for Rust project members to provide feedback on the Leadership Council. This is intended to be sent to `all@rust-lang.org`, and perhaps also post a link in #general.

cc @rust-lang/leadership-council

### DIFF
--- a/surveys/micro/leadership-council-2024.md
+++ b/surveys/micro/leadership-council-2024.md
@@ -1,0 +1,23 @@
+# Survey questions
+
+This is a quick annual survey for Rust-lang project members to provide feedback on the [Leadership Council](https://forge.rust-lang.org/governance/council.html). Your feedback is appreciated to help steer the Council activities. As always, if you have specific questions or feedback that you would like to provide, feel free to reach out to your Council representative.
+
+This survey is anonymous, though the Council will see the answers. A summary of these results will be released after the survey is complete.
+
+## Leadership Council Effectiveness
+
+### Do you feel that the Rust Leadership Council is serving its purpose effectively?
+
+Type: select one (optional)
+
+- Agree
+- Neither agree nor disagree
+- Disagree
+
+### Are there areas of the Leadership Council that you think are going well?
+
+Type: free text (optional)
+
+### Are there areas of the Leadership Council that you think could be improved or efforts that should be prioritized?
+
+Type: free text (optional)
--- a/surveys/micro/leadership-council-2024.md
+++ b/surveys/micro/leadership-council-2024.md
@@ -2,7 +2,7 @@
 
 This is a quick annual survey for Rust-lang project members to provide feedback on the [Leadership Council](https://forge.rust-lang.org/governance/council.html). Your feedback is appreciated to help steer the Council activities. As always, if you have specific questions or feedback that you would like to provide, feel free to reach out to your Council representative.
 
-This survey is anonymous, though the Council will see the answers. A summary of these results will be released after the survey is complete.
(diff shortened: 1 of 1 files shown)


Kobzol:
Remark: generally, it is easier to analyze closed-set answers, although open answers might provide more insights, of course. If this is targeted only for the Project members (which should be maybe 100 - 300 people? depending on how do we count), then it's not that terrible, but if you can think of interesting questions where e.g. a matrix answer could be used (I think that the council is X good in activity Y), that could make it easier to analyze the results and reduce the count of open answers.

apiraino:
I think I'd throw in maybe a couple of questions about the Council itself as a structure. Example:
- [ ] Do you know what the Council _actually_ is and does? Yes, No (explain why), I don't know
- [ ] Do you feel represented by the Council? Yes, No (explain why), I don't know

Note: I think the #general stream on Zulip is open to anyone, i.e. also Rust-curious not contributing to the project.

ehuss:
> matrix answer could be used (I think that the council is X good in activity Y), that could make it easier to analyze the results and reduce the count of open answers.

I don't think I have any specific activities to ask about in that way. However, I can see how directed questions can guide what kind of information we would like to get answered. I just don't know of anything specific to gather.

I'm not concerned about the number of responses with open questions. I'm not expecting a large number (and if there is a large number, that seems like it would be a good thing).

rylev:
+1 to @apiraino's suggestions perhaps rephrased like so:

> I am aware of the role that the Leadership Council plays in the governance of the Rust Project.
> * Strongly Agree
> * Agree
> * Unsure
> * Disagree
> * Strongly Disagree

I would also add more general questions about the state of Rust governance:
> The Rust Project has a solid foundation of Project governance.
> * Strongly Agree
> * Agree
> * Unsure
> * Disagree
> * Strongly Disagree

ehuss:
Thanks @rylev for the suggestions! I went ahead and added those.

@Kobzol The council discussed this together and I think we are ready to go.

eholk (approved):

surveys/micro/leadership-council-2024.md
```diff
@@ -0,0 +1,45 @@
+# Survey questions
+
+This is a quick annual survey for Rust-lang project members to provide feedback on the [Leadership Council](https://forge.rust-lang.org/governance/council.html). Your feedback is appreciated to help steer the Council activities. As always, if you have specific questions or feedback that you would like to provide, feel free to reach out to your Council representative.
+
+This survey is anonymous, though the Council will see the answers. A summary of these results will be released after the survey is complete.
```
Kobzol, line 5: > This survey is anonymous, though the Council will see the answers

This wording seems a bit weird. Of course that the Council will see the answers (that sounds obvious), but this wording makes it sound to me like the Council will see *who* answered (since the second part of the sentence is tied to the anonymity remark).

ehuss: Happy for any suggestions to make that clearer. I think there were two points that seemed important to me: it is anonymous, and who will see the answers. Unless stated, I don't think it is obvious who will see the answers since an alternate solution would be to have a third party summarize the results and the Council only sees a summary.

If it would help, we could drop the word `though` and just make it two sentences. Maybe something like:

> This survey is anonymous. The Council will go through all answers and release a summary after the survey is complete.

Kobzol: I think that's much better, thanks!

Kobzol:
Thanks! Prepared the survey and sent it to you via DM.

PR #265, opened 2024-07-26
Update micro-surveys description

Kobzol:
The five question limit was needlessly limiting, in my opinion.


apiraino (approved):
I agree, let's stay a bit more generic. The gist is that a micro-survey should be quickly answered and focus on a single question/topic.

PR #234, opened 2023-09-27
Bootstrap questions for the 2023 survey

Kobzol:
I copy pasted the 2022 questions and modified some small details (mostly Rust versions and dates), based on the previous diff from the 2021 to the 2022 surveys.

Then I went through the ~20 newest opened issues on this repository and tried to incorporate some "low hanging fruit" into the 2023 survey. I did the changes in individual commits, so that we can discuss and cherry pick them separately.

PR #273, opened 2024-09-06
Remove question about how often do people do coding/design work

Kobzol:
This question was originally suggested [here](https://github.com/rust-lang/surveys/pull/234/files#r1347513327).

While I think that the idea wasn't bad, I don't think that this question currently pulls its weight, so I'd like to remove it to shorten the survey. I'm basing this on the following two reasons:
1) This question only makes sense if we would do some complex cohort analysis and separate the answers of individual people. However, while SurveyHero can do this partially, we don't have access to it, and there's currently no automation (frankly, nor anyone who would have the time/motivation) that would help us with it. And without such analysis, this question is IMO not very useful.
2) Last year, we got the following answers:

In short, ~95 % people answered that they do coding or design work daily or weekly. I think that this shows the structure of our audience, that is probably highly technical. I think that we don't need to repeat the same question next time; I don't expect that it will change a lot.

CC @alice-i-cecile

### DIFF
--- a/surveys/2024-annual-survey/questions.md
+++ b/surveys/2024-annual-survey/questions.md
@@ -94,19 +94,6 @@ Type: select one
 > those who program once a week but always in Rust and those who program daily but
 > use Rust once a week.
 
-### On average, how often do you do coding/design work in general?
-
-Type: select one
-
-- Daily or nearly so
-- Weekly or nearly so
-- Monthly or nearly so
-- Rarely
-
-> **justification**
->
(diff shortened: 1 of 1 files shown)


Kobzol:
Regarding point 2, we also have the "Do you write or design software in your work?" question later in the survey.

alice-i-cecile:
Agreed, I don't think this is particularly valuable information to gather.

alice-i-cecile (approved):

PR #275, opened 2024-09-06
Move editor question after OS question

Kobzol:
I think it makes more sense there. Before, we asked about OS, then about Rust versions, then about IDE, and then about Rust features (which are related to used Rust version). I think that the new ordering makes more sense.

### DIFF
--- a/surveys/2024-annual-survey/questions.md
+++ b/surveys/2024-annual-survey/questions.md
@@ -206,6 +206,33 @@ Type: select all that apply (optional)
 > We specifically care about the runtime environment being targeted. ISA and other machine specifics are
 > not what matters.
 
+### Which editor or IDE setup do you use with Rust code on a regular basis?
+
+Type: select all that apply (optional)
+
+- VS Code
+- vi/vim/neovim
+- IntelliJ/CLion/other JetBrains IDE + Rust plugin
+- Rust Rover (dedicated IntelliJ Rust IDE)
+- Emacs (or derivatives like Doom Emacs, Spacemacs, etc.)
+- Sublime Text
+- Visual Studio
+- Xcode
+- Atom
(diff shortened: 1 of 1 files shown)


llogiq:
Helix and Lapce are missing.

Kobzol:
Huh, that's weird, Helix was actually in the survey, probably we monked patched it without including it in the repo, thanks.

Then there were a few open answers for Zed and Lapce, yes. However, I'm not sure where to draw the line for including it as an prepared answer. For example, Helix had 8% last year, which amounted to something like ~800 answers. However, in the free form answer field, Zed and Lapce had both only 70 and 50 answers, respectively.

Of course, the free form fields always receive much less answers than prepared answers, so maybe if Zed and Lapce were in the list, they would get much more?

Currently, we have 10 editors + open answer. I suppose that 12 editors wouldn't be that terrible, but we also shouldn't grow the list indefinitely.

apiraino:
I think every new option adds effort to the person filling the form, so let's be parsimonious :slightly_smiling_face: 

I also think the current panorama of tools still reflects year 2024. Adding Helix makes sense (given last years' results) but I'm not sure yet about "promoting" new ones out of the `other` option. If we notice an outlier that particularly stands out, that could be mentioned in the survey results comment.

Kobzol:
We wrote this last time:
image
which is good enough, I think.

We can observe the trends; last year, Zed got 70 answers, when it was macOS only. Now it also works on Linux, so I expect that it will be more popular. If it gets to a 100+ open answers this year, we can add it in the next survey.

Kobzol:
I will take a look how long is the current list in SurveyHero. If it looks like two other options are fine, I'll perhaps also add Lapce and Zed.

Issue #248, opened 2023-12-18
Improve question about stabilizing Rust features


Kobzol:
The question "Do you think that unimplemented (or nightly only) features should be stabilised in Rust?" has two answers:
- I think that unimplemented features should be stabilised
- I wish the Rust project to not add major new features (or slow down the pace of development)

But this question conflates two things:
- "I want some new features" vs "I don't want any new features"
- "New features should be added to Rust" vs "existing nightly features should be stabilized".

We should either reword or split the question in the next survey. [This Zulip topic](https://rust-lang.zulipchat.com/#narrow/stream/402479-t-community.2Frust-survey/topic/2023.20Survey.20Feedback/near/408645264) contains some suggestions.

PR #278, opened 2024-09-06
Ask about user's opinion about the speed with which Rust evolves

Kobzol:
The previous question was just... bad. This variation is hopefully more clear. I also think that it makes sense to ask this as the first question in this segment about Rust features.

Fixes: https://github.com/rust-lang/surveys/issues/248

### DIFF
--- a/surveys/2024-annual-survey/questions.md
+++ b/surveys/2024-annual-survey/questions.md
@@ -309,6 +309,23 @@ Type: select one (optional)
 > To get real data on how many people use older versions of the toolchain to
 > inform discussion on MSRVs.
 
+### What is your opinion on how fast Rust evolves?
+
+We want to know how do you perceive the speed of development of the Rust language.
+
+Type: select one (optional)
+
+- Rust is already too complex, it should not add or stabilize more significant features.
+- Rust changes too quickly, I wish it would slow down the pace of development.
+- I am satisfied with the current pace of development.
+- Rust changes too slowly, I wish it would add or stabilize features faster.
+- I don't know or don't care.
+
+> **justification**
+>
+> We want to find out if people prefer stability and fewer changes, or if they want to see more features being
+> implemented or stabilized.
+
 ### In your opinion, how would you prioritise work on the following aspects of Rust?
 
 Type: matrix (optional)
@@ -340,17 +357,6 @@ Priority:
 > **justification**
 >
 > This question gathers data on the communities perceptions of certain aspects of Rust at this point in time.
-
-### Do you think that unimplemented (or nightly only) features should be stabilised in Rust?
-
-Type: select one (optional)
-
-- I think that unimplemented features should be stabilised [`NEXT`](#which-unimplemented-or-nightly-only-features-are-you-looking-for-to-be-stabilized)
-- I wish the Rust project to not add major new features (or slow down the pace of development) [`NEXT`](#which-of-these-problems-do-you-recall-encountering-within-the-last-year)
-
-> **justification**
-> 
-> We want to find out if people prefer stability and fewer changes, or if they want to see more features being stabilized.  
 
 ### Which unimplemented or (nightly only) features are you looking for to be stabilized?
 


llogiq:
I fear that this question might be a bit too broad to be useful. E.g. I'd wish Rust would evolve faster in *some* directions, but slower in others... (Though given the question as is, I'd likely answer that I'm happy with the current pace).

What are we measuring here and what actionable learning do we expect from this?

Kobzol:
Good point, but I'm not sure if we can design a question that would enable us to find out the "pace opinion" for multiple areas (which areas?). Probably people will always want performance improvements, bug fixes etc., but might not want new features.

Regarding the motivation of this question, it's a bit complicated. It originally started with a question that follows this one, about features that people would like to see stabilized. This question was a bit controversial last year (see discussion in https://github.com/rust-lang/surveys/pull/243 if you're interested). What we would really like to find out are areas that Rust users struggle with, but it's not easy to figure that out with a closed answer set, and analyzing thousands of very liberally written open answers is also not something that we are good at. So we decided (as a compromise) to just include some basic list of most commonly requested features in the following question (and then we also have a question about problematic areas).

Then, out of this question, came up the idea about figuring out if people want *any* Rust features to be added at all. So what we are essentially measuring is the vibe about how fast Rust changes, if people want us to add new features or not, and how much has our complexity budget has already been spent. I think that this can be used to judge whether some new language features (which might be good ideas on its own) might perhaps be just too big of a change (like the recent `is` pattern matching discussion, or also the effect system comes to mind).

I think that the current set of questions about features/problems/pace of development is not great, but so far I have been struggling to figure out a better way to combine them together and derive what we actually want to ask our users.

surveys/2024-annual-survey/questions.md
```diff
@@ -290,6 +290,23 @@ Type: select all that apply (optional)
...
+
+We want to know if you are able to keep up with the speed of development of the Rust language.
+
+Type: select one (optional)
+
+- Rust is already too complex, it should not add or stabilize any more features.
+- Rust changes too quickly, I wish it would slow down the pace of development.
+- I am satisfied with the current pace of development.
+- Rust changes too slowly, I wish it would add or stabilize features faster.
+- I don't know or don't care.
```
apiraino, line 303: What is the problem you're actually trying to solve here? Why were the previous options "bad"? Do you want more nuance? Or you just wanted a better question wording? I am not fully sold on these new set of options, I feel they are too close to the extreme boundaries of the spectrum of what a user can think.

- "Rust is already too complex, it should not add or stabilize any more features".
  Comment: seems to imply that Rust should just ... not evolve anymore? Like a legacy programming language (i.e. with existing codebases but no new code being written)?

- "Rust changes too slowly, I wish it would add or stabilize features faster".
  Comment: "_Pull requests welcome_"?

To my eyes, these options seem to invite people to express entitled opinions about the development of Rust, almost like in a employer-employee work relation. Are we working hard enough for you? :laughing:

Kobzol: So, first let me say that the question that we had last year just.. sucked; I think that was kind of universally agreed upon, both from feedback on Reddit and also on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/402479-t-community.2Frust-survey/topic/2023.20Survey.20Feedback/near/408645264).

The original motivation for the question, as suggested by Jyn, was to get a vibe check about the speed with which Rust evolves. However, I worded the question very poorly:
```
I think that unimplemented features should be stabilised
vs
I wish the Rust project to not add major new features (or slow down the pace of development)
```

We asked if the user thinks that unimplemented features should be stabilised, which is just wrong for several reasons (is unimplemented = unstable, or just missing at all? why shouldn't they get stabilized? this probably depends on each specific feature? etc.). And the answers also presented a false dichotomy. And they missed a "I don't know" option.

So my first intention is to just nuke the question, as it was just bad.

The second intention is to replace it with a better question. I agree that the answers contain also the extreme sides of the spectrum, that was deliberate. I find it a bit baffling, but from what I've seen online, there are indeed voices that claim that Rust should just stop adding features, period, and just focus on polishing stuff. I find this view a bit too extreme, but it's certainly a valid opinion. Another valid opinion/request is people wanting Rust to evolve faster. They cannot really make any demands towards that, and as you said, if they want to help, they should start contributing, but that's true universally.

In the survey, we ask what is the users opinion about Rust; we don't promise anything. On the other hand, I think that it is quite interesting how do people see Rust's "change cadence". It probably won't happen, but if for some reason 95% people told us that they don't want tnew features, then as weird as it may sound, I think that it should be taken as an input for further development of Rust (again, I don't think that this will happen).

surveys/2024-annual-survey/questions.md
```diff
@@ -290,6 +290,23 @@ Type: select all that apply (optional)
 > in a separate question, though this is likely easier to find out through download
 > numbers.
 
+### What is your opinion on how fast Rust evolves?
+
+We want to know if you are able to keep up with the speed of development of the Rust language.
```
apiraino, line 295: Also, this wording puts implicitedly the blame on the user (_"are you good enough for Rust?"_). I would invert the polarity with something like "we want to know if the Rust PL development is moving too fast"

surveys/2024-annual-survey/questions.md
apiraino: Hmm ... I understand your overall reasoning but at the same time I can't parse it quite well :slightly_smiling_face: 

So, what are we looking for: feedback about the pace of development, right? I think one fundamental data point we gathered in 2023 survey was that the **No. 1 voted option** for "what are your biggest worries about Rust" was "becomes too complex". The way I see it, people are literaly screaming at us saying "Rust is a tough nut to crack".

This new question looks almost like a duplicate or a second, expanded version of "what are your biggest worries about Rust". What we _could_ ask is instead if how people feel about us working more on making Rust "easier" (blanket word for multiple concepts) or add "new stuff". The problem is that they intersect. To make things "easier" we sometimes need to add "new stuff".

surveys/2024-annual-survey/questions.md
Kobzol: Good point!

surveys/2024-annual-survey/questions.md
Kobzol: You're right on both accounts :) This is essentially a questions that "zooms-in" on the "becomes too complex" opinions. I don't see that as a bad thing, but I also don't have a strong opinion about it, I'm fine with just removing this question.

Regarding the second point: indeed there are some features that are more about "filling in gaps in the existing system" rather than "adding new features". For example, async fn in traits is essentially enabling something that people intuitively expected to work before. That's quite different from e.g. adding an effect system that could transform how Rust is written  in a large way.

However, as usually with these questions, the devil is in the details; it's hard to convey all these little nuances in a few words.. :)

Kobzol:
Rebased and modified the perex + the first answer.

apiraino (approved):
OK, I'm still not completely sold to this question but let's try asking either way, let's see if interesting data comes out :)

surveys/2024-annual-survey/questions.md
apiraino: > if for some reason 95% people told us that they don't want new features (...) (I don't think that this will happen)

if _that_ happens I'm outta here LOL

Issue #253, opened 2023-12-19
Ask about stdlib features


Kobzol:
Our questions about missing features doesn't specifically mention stdlib (e.g. SIMD etc.). It would be nice to add it. Noted on [Reddit](https://www.reddit.com/r/rust/comments/18l7ryc/comment/ke012qy).

apiraino:
Added also a screenshot for archiving purposes

<details><summary>Details</summary>
<p>

![grafik](https://github.com/rust-lang/surveys/assets/6098822/220c1a47-1c4c-4369-bfbb-7b82b0b556fb)


</p>
</details>

apiraino:
Asked t-libs on [Zulip ](https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/Rust.20survey.202024)

PR #279, opened 2024-09-06
Expand question about which features should be stabilized

Kobzol:
I included more features, and changed the answers, because I felt like high/medium/low priority is both hard to determine and does not tell us the useful information (which features are wanted by people, which are just nice to have, which are not wanted that much and which are not even known to people).

Fixes: https://github.com/rust-lang/surveys/issues/253

### DIFF
--- a/surveys/2024-annual-survey/questions.md
+++ b/surveys/2024-annual-survey/questions.md
@@ -368,18 +368,19 @@ Features:
 
 - Specialization ([RFC#1210](https://github.com/rust-lang/rust/issues/31844))
 - Generators/coroutines ([RFC#2033](https://github.com/rust-lang/rust/issues/43122))
+- Async closures ([RFC#2532](https://github.com/rust-lang/rust/issues/62290))
 - Async generators/coroutines (https://github.com/rust-lang/rust/pull/118420)
 - If/while let chains ([eRFC#2497](https://github.com/rust-lang/rust/issues/53667))
 - Try blocks ([RFC#243](https://github.com/rust-lang/rust/issues/31436))
 - Never type ([RFC#1216](https://github.com/rust-lang/rust/issues/35121))
-- Trait aliases
-- Type alias impl trait (TAIT)
-- Associated type defaults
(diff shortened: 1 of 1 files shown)


surveys/2024-annual-survey/questions.md
```diff
@@ -333,33 +333,39 @@ Type: select one (optional)
...
 Priority:
 
-- High Priority
-- Medium Priority
-- Low Priority
-- Should not be prioritised
+- Need this
+- Nice to have
+- Don't need this
+- Don't know what it is
```
Noratrieb, line 368: oh these are great, nice

Noratrieb: although actually no, I don't like this. I would focus more on "this solves a blocker I have", "this solves a problem I have", "this does not solve a problem I have" (can be reworded a bit better). people often *need* a specific feature, but a specific feature might resolve a blocker they have (that could maybe also be resolved differently)

apiraino: The new "priority" options are sound better than last year, but I'd prefer them to be worded in a way to avoid that slight bitter taste when reading them (see [my comment](https://github.com/rust-lang/surveys/pull/278#discussion_r1749909270) in a related topic).

I like the wording of Nora.

Kobzol: I was quite clueless on designing these answers, I just knew that the priority answers were bad :laughing: So I'm open to any suggestions. What about something like this?

```
- Would unblock my use-case
- Would improve my code
- Don't need it
- Don't know what it is
```

Btw, I'm not sure if we should even keep this question (see discussion in https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/Rust.20survey.202024). Maybe it would be really better to focus on the problems that users have, rather than on the features.

apiraino (approved):
I think all the points in this discussion have at least a bit of truth. In the end I think leaving this question in the survey won't be bad.

PR #281, opened 2024-09-07
Split macOS into x64 and ARM variants

Kobzol:
This would tell us more about which of those two common hardware platforms are used the most on macOS. Suggested by @lqd.

### DIFF
--- a/surveys/2024-annual-survey/questions.md
+++ b/surveys/2024-annual-survey/questions.md
@@ -161,7 +161,8 @@ Type: select all that apply (optional)
 - Windows 10/11
 - Windows 8 or older
 - Windows Subsystem for Linux
-- macOS
+- macOS (Intel)
+- macOS (ARM)
 - Other (open response)
 
 > **justification**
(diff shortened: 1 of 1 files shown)


surveys/2024-annual-survey/questions.md
```diff
@@ -161,7 +161,8 @@ Type: select all that apply (optional)
 - Windows 10/11
 - Windows 8 or older
 - Windows Subsystem for Linux
-- macOS
+- macOS (x64)
```
Noratrieb, line 164: does apple call them like that too? I thought "x64" was mostly a Windows thing.
but either way I think this should elaborate and probably say something like "macOS (Intel)", "macOS (M-series Apple Silicon)" or something like that, to make it clear what this means. not everyone is familiar with the technical name of their CPU architecture

Kobzol: Yeah, I was a bit lazy with the name :laughing: Changed it to `Intel`, that should be universally understandable, I hope.

lqd (approved):
👍

lqd:
Although, nora's suggestion of using macOS M-series or Apple Silicon may also be a similar improvement as switching from x64 to Intel.

PR #283, opened 2024-09-15
Clarify answers for Rust job marker for junior/senior role

Kobzol:
There's not much that we can do with the answers of this question, but this seemed to me like a tiny change that could make the answers a bit more interesting.

No strong feeling about this though, I'm also fine to just keep it as it is.


apiraino (approved):
Basically " :man_shrugging: " from me. I don't see a lot of value splitting this into 2 options but I'm not against it either.

PR #284, opened 2024-09-26
Reword and merge 2 questions about Rust desiderata

apiraino:
This is a suggested merge of two similar questions about what people would like to see improved in the Rust project.

I merged these 2 questions:
- [Which of the following aspects of Rust present non-trivial problems to your programming productivity](https://github.com/rust-lang/surveys/blob/f48625bf2b4c99efc3977c8d237ae4b8e61bfcca/surveys/2024-annual-survey/questions.md?plain=1#L404-L417)
- [In your opinion, how would you prioritise work on the following aspects of Rust?](https://github.com/rust-lang/surveys/blob/f48625bf2b4c99efc3977c8d237ae4b8e61bfcca/surveys/2024-annual-survey/questions.md?plain=1#L329C1-L355)

Eliminating the latter and changing the format of the former, now a multiple choice matrix where people can express how important is each feature for them.

@Kobzol we can now iterate on the options and remove what we think is redundant or not useful

### DIFF
--- a/surveys/2024-annual-survey/questions.md
+++ b/surveys/2024-annual-survey/questions.md
@@ -326,38 +326,6 @@ Type: select one (optional)
 > We want to find out if people prefer stability and fewer changes, or if they want to see more features being
 > implemented or stabilized.
 
-### In your opinion, how would you prioritise work on the following aspects of Rust?
-
-Type: matrix (optional)
-
-Aspects:
-
-- Runtime performance
-- Compile times
-- Binary size
-- Memory usage (i.e., how much RAM rustc uses when compiling)
-- Disk space usage (e.g., the size of target folder)
-- Bugs in the compiler (i.e., ICEs a.k.a. internal compiler errors, miscompilations, etc.)
-- Compiler error messages
-- IDE experience
-- Debugging experience
-- Documentation (rustdoc, docs.rs)
-- Build system (cargo)
-- Package management (crates.io)
-- New language features
-- Rust language and standard library documentation
-
-Priority:
-
-- High Priority
-- Medium Priority
-- Low Priority
-- Should not be prioritised
-
-> **justification**
->
-> This question gathers data on the communities perceptions of certain aspects of Rust at this point in time.
-
 ### Which unimplemented or (nightly only) features are you looking for to be stabilized?
 
 Please mention *Rust compiler or standard library* features that are currently unstable (only available using a nightly release of the Rust compiler) or missing that would in your opinion be beneficial to the Rust ecosystem or to you. This list excludes other tooling around the compiler such as cargo, rustup, rustfmt, etc.
@@ -403,22 +371,47 @@ Type: free text
 
 ### Which of the following aspects of Rust present non-trivial problems to your programming productivity?
 
-Type: select all that apply (optional)
+#### Choose an assessment only for challenges that you think are impacting your work. You can skip any topic, we will interpret that as you are neutral.
 
-- Implementing things on tuples
-- Splitting things across crates (orphan rule)
-- Having to do iterator implementations manually
+Type: matrix (optional)
+
+- Implementing logic for tuples of various sizes
+- Splitting code across crates (orphan rule)
+- Having to implement Iterator manually
 - Not being able to do enough in const fn
-- Implementing Rust plugins with a stable ABI
-- Interoperating with other languages (e.g. C and C++)
+- Implementing plugins
+- Interoperating with other languages (e.g. C or C++)
 - Achieving structured concurrency with async code
 - Writing executor-agnostic async code
+- Writing correct unsafe code
 - Borrow checker not allowing valid code
-- Other (please specify)
+- Slow runtime performance
+- Slow compile times
+- Large binary size of compiled artifacts
+- High disk space usage (e.g. the size of the target folder)
+- Encountering compiler bugs (e.g. ICEs a.k.a. internal compiler errors or miscompilations)
+- Encountering opaque/unclear compiler error messages
+- Subpar IDE support (e.g. some errors are not shown or analysis is slow)
+- Subpar debugging experience (e.g. missing value visualizations or async stacktraces)
+- Lacking documentation of the Rust language or the standard library
+
+Options:
+
+- Big problem for me
+- Could be improved, but does not limit me
+- Not an issue for me at all
 
 > **justification**
 >
-> Conversely, try asking which /problems/ they encountered and let the Rust project figure out what is needed to improve in these areas
+> This question tries to get insights on what people wish the Rust project would improve.
+
+### Are there any challenges not mentioned above that limit your Rust programming productivity?
+
+Type: free text
+
+> **justification**
+>
+> Allow the cohort to mention other challenges.
 
 ### Which features stabilized in the past 12 months do you use the most?
 
--- a/surveys/2024-annual-survey/questions.md
+++ b/surveys/2024-annual-survey/questions.md
@@ -371,7 +371,7 @@ Type: free text
 
 ### Which of the following aspects of Rust present non-trivial problems to your programming productivity?
 
-#### Choose an assessment only for challenges that you think are impacting your work. You can skip any topic, we will interpret that as you are neutral.
+#### Assess only challenges that you think are impacting your work.
 
 Type: matrix (optional)
 
--- a/surveys/2024-annual-survey/questions.md
+++ b/surveys/2024-annual-survey/questions.md
@@ -405,7 +405,7 @@ Options:
 >
 > This question tries to get insights on what people wish the Rust project would improve.
 
-### Are there any challenges not mentioned above that limit your Rust programming productivity?
+### Are there any challenges not mentioned above that affect your Rust programming productivity?
 
 Type: free text
 


surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
...
+
+- (From here the additional one)
+- Runtime performance
+- Compile times
+- Binary size (of the final compiled artifact)
+- Memory usage (example: how much RAM rustc uses when compiling)
+- Disk space usage (example: the size of target folder)
+- Bugs in the compiler (that is, ICEs a.k.a. internal compiler errors, miscompilations, etc.)
+- Opaque / unclear compiler error messages
+- IDE experience
```
apiraino, line 394: IDE is a bit vague (I use emacs, you use vstudio). If this question means rust analyzer maybe explicit it?

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
...
+- Debugging experience (example: unclear / misleading diagnostics)
+- Documentation (rustdoc, docs.rs)
+- Build system (cargo)
+- Package management (crates.io)
+- New language features
```
apiraino, line 399: This is also vague ("I want new things" - yeah but what?). I think we can remove it.

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
 - Achieving structured concurrency with async code
 - Writing executor-agnostic async code
 - Borrow checker not allowing valid code
+
+- (From here the additional one)
+- Runtime performance
```
apiraino, line 387: This is also a blanket statement I think it's not very useful. Does it make sense to say "faster!!11" without specifying where/when/what?

Kobzol: I think that we can reword everything to be a "challenge" or a "problem" (since we're looking for problems and roadblocks, rather than solutions). So having "slow runtime performance" seems fine to me.

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
...
+- Compile times
```
Kobzol, line 388: In the same vein, I'd write something like "Compilation is too slow" (we could even split it into incremental vs from scratch builds).

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
...
+- Binary size (of the final compiled artifact)
```
Kobzol, line 389: "Achieving small binary size of compiled artifacts" or "Rust binaries are too large".

surveys/2024-annual-survey/questions.md
Kobzol: I would reword to something like "Insufficient IDE support" or "Rust IDE support is not good enough or is too slow", or something like that.

surveys/2024-annual-survey/questions.md
Kobzol: Yeah, I would remove it, along with the blanket statements, like documentation or build system. It should be specific things, like "Code/feature that I'm using lacks documentation" or something like that.

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
...
+- Package management (crates.io)
+- New language features
+- Rust language and standard library documentation
 - Other (please specify)
 
+How bad do I need this?:
+
+- Would unblock my use-case
+- Would greatly improve my developer experience
+- Not a priority / good enough for me
```
Kobzol, line 407: We should also add an option I don't know (or I don't care), and clarify that not choosing an answer equals to saying "I don't know".

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
...
+- Bugs in the compiler (that is, ICEs a.k.a. internal compiler errors, miscompilations, etc.)
+- Opaque / unclear compiler error messages
+- IDE experience
+- Debugging experience (example: unclear / misleading diagnostics)
+- Documentation (rustdoc, docs.rs)
+- Build system (cargo)
+- Package management (crates.io)
+- New language features
+- Rust language and standard library documentation
 - Other (please specify)
```
Kobzol, line 401: Matrix questions in SH sadly cannot have the "Other" open category :( So we will need additional question anyway, to ask for open answers. This would mean that we wouldn't reduce question count by one, which is a bit annoying, but nothing too bad I guess.

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
...
+- (From here the additional one)
+- Runtime performance
+- Compile times
+- Binary size (of the final compiled artifact)
+- Memory usage (example: how much RAM rustc uses when compiling)
+- Disk space usage (example: the size of target folder)
+- Bugs in the compiler (that is, ICEs a.k.a. internal compiler errors, miscompilations, etc.)
+- Opaque / unclear compiler error messages
+- IDE experience
+- Debugging experience (example: unclear / misleading diagnostics)
```
Kobzol, line 395: Debugging experience is not good enough.

We could even add another one for async debugging, but that's probably too detailed (and we know anyway that async debugging in Rust sucks now, so.. :) ).

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
...
+- Documentation (rustdoc, docs.rs)
+- Build system (cargo)
```
Kobzol, line 397: Building Rust projects?

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
...
+- Package management (crates.io)
```
Kobzol, line 398: Using dependencies? Or e.g. auditing dependencies? We can just arbitrarily choose some common complaints :)

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
 - Achieving structured concurrency with async code
 - Writing executor-agnostic async code
 - Borrow checker not allowing valid code
+
+- (From here the additional one)
+- Runtime performance
+- Compile times
+- Binary size (of the final compiled artifact)
+- Memory usage (example: how much RAM rustc uses when compiling)
+- Disk space usage (example: the size of target folder)
```
Kobzol, line 391: `Disk space usage is too high"?

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,33 @@ Type: select all that apply (optional)
 - Achieving structured concurrency with async code
 - Writing executor-agnostic async code
 - Borrow checker not allowing valid code
+
+- (From here the additional one)
+- Runtime performance
+- Compile times
+- Binary size (of the final compiled artifact)
+- Memory usage (example: how much RAM rustc uses when compiling)
```
Kobzol, line 390: Memory usage of the Rust compiler or the compiled Rust programs.

Kobzol (commented):
Thanks! I think this combination is good. Once we figure out the options, we should do a final-ish pass and reorder them so that they are somewhat logically grouped together.

Kobzol:
I also asked some people IRL on a Rust Meetup yesterday, and they suggested a few additional points:
- Expressing bounds on async fns in traits (but this is already mostly solved in nightly, so...)
- Choosing the right string type
- Refactoring (adding/removing lifetimes, generic parameters)
- Code readability

Just throwing this out here :)

surveys/2024-annual-survey/questions.md
apiraino: I would keep the wording neutral because the quality assessment (good/bad) is up to the surveyer.

I was just wondering if it makes sense to coalesce all IDEs together, each of them has a completely different story and lifecycle. Maybe I use X and it's a stellar experience and you use Y and it's bad, it's apple and oranges and these differences are lost here.

Well, let's leave it like that, there's a specific question about "which IDE do you use" :man_shrugging:

surveys/2024-annual-survey/questions.md
apiraino: Heh, this is also a wide topic. I don't know how to change this, what do we really want to point at

surveys/2024-annual-survey/questions.md
apiraino: :shrug:

surveys/2024-annual-survey/questions.md
Kobzol: Or we can just scrap this. I just wanted to avoid the situation from last year, where we asked "which Rust aspects are causing you issues" and one of the answers was just "async" or "macros" xD

surveys/2024-annual-survey/questions.md
apiraino: but these are two different questions, right?

surveys/2024-annual-survey/questions.md
```diff
@@ -414,11 +382,28 @@ Type: select all that apply (optional)
...
+- Disk space usage (example: the size of target folder)
+- Bugs in the compiler (that is, ICEs a.k.a. internal compiler errors, miscompilations, etc.)
+- Opaque / unclear compiler error messages
+- Subpar IDE support
+- Debugging experience is not good enough (example: unclear / misleading diagnostics)
+- Package management (crates.io)
+- Rust language and standard library documentation
+
+How bad do I need this?:
```
Kobzol, line 397: There isn't actually any label for this in SurveyHero, so this text ("How bad do I need this") won't appear anywhere. Which is good, because the wording would be weird ("Slow compile times" -> "How bad do I need this" is probably not a good lead-in :laughing: ). We could add some text just below the question text though.

So either we have to ask the main question in a way that the prepared answers make sense for it, or explain the answers below the question.

It seems to me that the answers in the vein of "it would do something for me" are too.. complicated, so that they make sense across all the options. What about something like this? It's hard to come up with good options for all four columns though..
image

apiraino: How does SH recognize and separate in the matrix the block of answer options from the topics? "How bad do I need this" was just a funny block separator between the two but it can be removed (and def. should be reworded).

Choose an assessment only for the topics you think they are impacting your work. If you don't choose an answer, we will interpret as "I don't know" or that you are neutral on that topic.

About the column layout: now that I see it in the layout I'd say let's remove the "I don't know" option to win some horizontal space an try to reword the answer options as you suggest to keep them as short as possible.

Kobzol: > How does SH recognize and separate in the matrix the block of answer options from the topics? "How bad do I need this" was just a funny block separator between the two but it can be removed (and def. should be reworded).

It doesn't, we fill in the questions into SH manually xD We do tend to include *some* separator in the Markdown file though, to recognize the split, as you said.

> About the column layout: now that I see it in the layout I'd say let's remove the "I don't know" option to win some horizontal space an try to reword the answer options as you suggest to keep them as short as possible.

Ok, we can just interpret a missing answer as "I don't know" (as the text already says), it will just need a bit of postprocessing.

apiraino: I have pushed a proposal to drastically shorten the colum width? How does it look?

Kobzol: Having just three columns is :+1: But yeah, I'll need to think of the wording to unify it. "slow runtime performance" -> "very much needed" isn't ideal xD I'll try to rephrase all questions and the answers so that it makes more sense.

apiraino: > I'll try to rephrase all questions and the answers so that it makes more sense.

I'll also have a look at that

Kobzol:
Suggestion:
```
### Which of the following challenges present non-trivial problems to your Rust programming productivity?

Rate how much do these challenges affect your productivity.
Please rate only challenges that actually impact your work.
You can skip any challenge, we will interpret that as answering "I don't know or don't care".

Type: matrix (optional)

- Implementing logic for tuples of various sizes
- Splitting code across crates (orphan rule)
- Having to implement Iterator manually
- Not being able to do enough in const fn
- Implementing plugins
- Interoperating with other languages (e.g. C or C++)
- Achieving structured concurrency with async code
- Writing executor-agnostic async code
- Borrow checker not allowing valid code
- Slow runtime performance
- Slow compile times
- Large binary size of compiled artifacts
- High disk space usage (e.g. the size of the target folder)
- Encountering compiler bugs (e.g. ICEs a.k.a. internal compiler errors or miscompilations)
- Encountering opaque/unclear compiler error messages
- Subpar IDE support (e.g. some errors are not shown or analysis is slow)
- Subpar debugging experience (e.g. missing value visualizations or async stacktraces)
- Lacking documentation of the Rust language or the standard library

Options:

- Big problem for me
- Could be improved, but does not limit me
- Not an issue for me at all

> **justification**
>
> This question tries to get insights on what people wish the Rust project would improve.

### Are there any challenges not mentioned above that limit your Rust programming productivity?

Type: free text

> **justification**
>
> Allow the cohort to mention other challenges.
```

surveys/2024-annual-survey/questions.md
```diff
@@ -403,22 +371,47 @@ Type: free text
 
 ### Which of the following aspects of Rust present non-trivial problems to your programming productivity?
 
-Type: select all that apply (optional)
+#### Choose an assessment only for challenges that you think are impacting your work. You can skip any topic, we will interpret that as you are neutral.
```
apiraino, line 374:
```suggestion
#### Assess only challenges that you think are impacting your work.
```

surveys/2024-annual-survey/questions.md
```diff
@@ -403,22 +371,47 @@ Type: free text
...
+
+- Big problem for me
+- Could be improved, but does not limit me
+- Not an issue for me at all
+
+> **justification**
+>
+> This question tries to get insights on what people wish the Rust project would improve.
+
+### Are there any challenges not mentioned above that limit your Rust programming productivity?
```
apiraino, line 408:
```suggestion
### Are there any challenges not mentioned above that affect your Rust programming productivity?
```

surveys/2024-annual-survey/questions.md
```diff
@@ -403,22 +371,47 @@ Type: free text
...
+- Encountering compiler bugs (e.g. ICEs a.k.a. internal compiler errors or miscompilations)
+- Encountering opaque/unclear compiler error messages
+- Subpar IDE support (e.g. some errors are not shown or analysis is slow)
+- Subpar debugging experience (e.g. missing value visualizations or async stacktraces)
+- Lacking documentation of the Rust language or the standard library
+
+Options:
+
+- Big problem for me
+- Could be improved, but does not limit me
```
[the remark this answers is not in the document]
Kobzol: It's severe bikeshedding at this point :laughing: But I think that this changes the meaning quite a lot.

The important message that I want to get out of this is whether some challenge is a limitation/blocker/big problem or rather "just an annoyance". This second answer should have been "just an annoyance".

Saying "it does not affect me" overlaps with the third option, and thus there's no option that would state "I see this as a problem, but it's not *that* bad". For example, there are things that probably affect almost everyone (like compile times), that everyone wants to get improved (who woud want *longer* compile times?), but for some people it's not really a blocker for their productivity.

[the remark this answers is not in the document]
apiraino: agreed. thanks for clarifying the point (which I had missed)

Kobzol:
Ok, I'm now much happier with these questions than before. Unless you want to change something, feel free to merge!

Kobzol (approved):

PR #302, opened 2024-11-14
Align rustc versions

apiraino:
This should bring in all translations the rustc versions to be the same for the question "which is the oldest rustc do you use"


Kobzol (approved):
Thanks!

PR #305, opened 2024-11-18
Slightly revise translation for `associated_type_bounds`

jieyouxu:
It should not have anything about `item`s.


Kobzol:
Just a note: when making changes to the Markdown, please always also make them in SurveyHero, I thought that you made the change there, but now I noticed that the original text is there. I removed the `item` substring from SH, hopefully that was the right modification :laughing:

jieyouxu:
Oh right I forgor, sorry!

PR #309, opened 2024-12-01
Add contributors guidelines

apiraino:
hey @Kobzol I was thinking to write here a document to help contributors orient themselves. Put here all the tribal knowledge we may have sparse on Zulip.

wdyt?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,29 @@
+# How to contribute to the State of Rust survey
+
+Great that you want to help with the State of Rust survey! The State of Rust survey is the annual platform for us to look at how was the past year of work and for the community to express their wishes.
+
+The survey platform of choice is currently [Survey Hero](https://www.surveyhero.com). We have reached a general consensus that, while not perfect, it works rather well for us therefore we are not looking into alternatives for the time being.
+
+The State of Rust survey is published usually in the first half of December. Contributions are tipically needed once a year as we approach to the deadline. The initial work to kickoff the survey is taken care of by the [Survey Team](https://github.com/rust-lang/team/blob/master/teams/community-survey.toml). Contributions are welcome to translate the survey into multiple languages. The more, the merrier.
+
+Here's a rough timeline of the work:
+- The State of Rust survey is discussed on Zulip (in [#t-community/survey](https://rust-lang.zulipchat.com/#narrow/channel/402479-t-community.2Frust-survey)). We gather feedback about the previous year's survey, we ask the main teams if they have specific wishes and we draft the English survey
+- In the meantime we start recording interest from volunteers to translate the survey into different languages
+- After the English survey is finalized and uploaded on Survey Hero, translators can start their work
+
+Here's how the translation effort is coordinated:
+- The Survey Team will first generate translation drafts by using a Survey Hero tool to machine-translate the English survey into other languages. The team will generate drafts only for languages we know will be reviewed by a human.
+- Contributors will Keep an eye on the Zulip topic and will be contacted to receive an access token to access Survey hero and be able to work on a specific translation.
+- Once the human review on Survey Hero is complete, the Survey Team will download the machine-translated drafts (Markdown files) and open pull requests on this repository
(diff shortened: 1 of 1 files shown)


Kobzol:
Isn't that already kind of this: https://github.com/rust-lang/surveys/blob/main/annual-survey.md?

apiraino:
> Isn't that already kind of this: [`main`/annual-survey.md](https://github.com/rust-lang/surveys/blob/main/annual-survey.md?rgh-link-date=2024-12-02T09%3A16%3A11Z)?

I read `annual-survey.md` more for the survey team (today: us). The CONTRIBUTING.md was meant for people interested to help with the translations.

But is your comment about having two similar documents or about the content my patch being redundant in general? In the former case, happy to merge both into one. In the latter case ... just close and discard this patch?

Kobzol:
I see, you're right. There is slight overlap, but I think that a contributing guide for non-survey-team members is still useful. The current version in this PR is mostly aimed at translations, maybe we could generalize it a bit to also tell people what to do when they want to modify also the main English version?

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
+# How to contribute to the State of Rust survey
+
+Great that you want to help with the State of Rust survey! The State of Rust survey is the annual platform for us to look at how was the past year of work and for the community to express their wishes.
+
+The survey platform of choice is currently [Survey Hero](https://www.surveyhero.com). We have reached a general consensus that, while not perfect, it works rather well for us therefore we are not looking into alternatives for the time being.
+
+The State of Rust survey is published usually in the first half of December. Contributions are tipically needed once a year as we approach to the deadline. The initial work to kickoff the survey is taken care of by the [Survey Team](https://github.com/rust-lang/team/blob/master/teams/community-survey.toml). Contributions are welcome to translate the survey into multiple languages. The more, the merrier.
```
Kobzol, line 7:
```suggestion
The State of Rust survey is published usually in the first half of December. Contributions are typically needed once a year as we approach the deadline. The initial work to kickoff the survey is taken care of by the [Survey Team](https://github.com/rust-lang/team/blob/master/teams/community-survey.toml). Contributions are welcome to translate the survey into multiple languages or to suggest changes to the contents of the survey. The more, the merrier.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+
+Here's a rough timeline of the work:
+- The State of Rust survey is discussed on Zulip (in [#t-community/survey](https://rust-lang.zulipchat.com/#narrow/channel/402479-t-community.2Frust-survey)). We gather feedback about the previous year's survey, we ask the main teams if they have specific wishes and we draft the English survey
+- In the meantime we start recording interest from volunteers to translate the survey into different languages
+- After the English survey is finalized and uploaded on Survey Hero, translators can start their work
+
+Here's how the translation effort is coordinated:
```
Kobzol, line 14: I'm not sure if we want to use this workflow. A large part of the help provided by the translators is that they have actually performed the translation directly in SurveyHero, which was quite nice. Especially for `zh-CN`, I couldn't even replicate the SH contents by copy-pasting from Markdown, lol.

On the other hand, reviewing the changes on GitHub is *much* easier, and multiple people can coordinate. And now that we actually have decent tooling, and hopefully also a reasonably good translation baseline on SH, it might not be so bad to just do everything on GH, and then do one final manual sync to SH. It was sure quite messy this year, when the changes were made to both SH and MD in parallel, and we had to frequently resync in both directions.

apiraino: You're right, the workflow I described is in part wrong (the first step is the contributor working on SH). So for now I have amended the description to reflect this years' workflow because it has happened (please check if now is correct).

How do you prefer to work next year? GH only? Or like this year? Or do we want to discuss this at a later time, after we're done?

Kobzol: The current description looks good. Let's discuss the rest next year :laughing:

Kobzol:
Thank you!

PR #313, opened 2024-12-02
Backport a few French changes from SurveyHero

Kobzol:
With this, I consider the translations to be finished.


apiraino:
Looks ok! thanks you jakob :pray:

PR #315, opened 2024-12-03
Last fixes

apiraino:
Let's put here all the small things we notice

r? @Kobzol

### DIFF
--- a/surveys/2024-annual-survey/translations/de.md
+++ b/surveys/2024-annual-survey/translations/de.md
@@ -669,7 +669,7 @@ Type: select one
 - < 10 Jahre
 - > 10 Jahre
 
-### Wo wohnen Sie?(Bitte entschuldigen Sie, dass die Auswahloptionen nicht korrekt sortiert sind)
+### Wo wohnen Sie?
 
 Type: select one
 
@@ -712,8 +712,8 @@ Type: select one
 - Dänemark
 - Ecuador
 - Ein anderes Land
-- Elfenbeinküste
 - El Salvador
+- Elfenbeinküste
 - Eritrea
 - Estland
 - Fidschi
@@ -860,8 +860,8 @@ Type: select one
 - Venezuela
 - Vereinigte Arabische Emirate
 - Vereinigte Republik Tansania
-- Vereinigtes Königreich Großbritannien und Nordirland
 - Vereinigte Staaten von Amerika
+- Vereinigtes Königreich Großbritannien und Nordirland
 - Vietnam
 - Weißrussland
 - Zentralafrikanische Republik
--- a/surveys/2024-annual-survey/questions.md
+++ b/surveys/2024-annual-survey/questions.md
@@ -1,6 +1,6 @@
 # Survey questions
 
-Whether or not you use Rust Programming Language <https://rust-lang.org> today, we want to hear from you!
+Whether or not you use the Rust Programming Language <https://rust-lang.org> today, we want to hear from you!
 
 The Rust Community Team has created this survey to help us gauge how we're doing, what can be improved, and how we can best engage with all of you as we move forward.
 


surveys/2024-annual-survey/translations/de.md
```diff
@@ -669,7 +669,7 @@ Type: select one
 - < 10 Jahre
 - > 10 Jahre
 
-### Wo wohnen Sie?(Bitte entschuldigen Sie, dass die Auswahloptionen nicht korrekt sortiert sind)
+### Wo wohnen Sie?
```
apiraino, line 672: It's not true that the countries are not sorted :)

Marcono1234: This hasn't been synced? The survey still shows this note.

Though the sorting is a bit problematic because "Other" (German "Ein anderes Land") also appears in sort order so it is quite likely that users will overlook it (same for English).

surveys/2024-annual-survey/translations/de.md
```diff
@@ -712,8 +712,8 @@ Type: select one
 - Dänemark
 - Ecuador
 - Ein anderes Land
-- Elfenbeinküste
 - El Salvador
+- Elfenbeinküste
```
Marcono1234, line 716: Note that this does _not_ match the actual sort order shown on the survey (same for "Vereinigtes Königreich ..."). The previous version was actually matching the sort order of the survey.

surveys/2024-annual-survey/translations/de.md
Kobzol: To clarify, we won't be making any changes to the survey this year (the survey team doesn't even have access to it anymore, but more importantly we don't want to make changes while the survey is running). The fixes will be applied next year.

Marcono1234: Do you mean this specific to this comment, or generally for all changes (e.g. also #322)?

https://blog.rust-lang.org/2024/12/05/annual-survey-2024-launch.html explicitly asks for pull requests for the translations.

Kobzol: I mean generally for all changes. Yes, we do ask for PRs, and are very glad for them! As they will improve the translation for next year(s). We never modify the survey while it is running, unless there is really something critical going on. We should perhaps clarify this in the blog post next time.

Marcono1234: > unless there is really something critical going on

Not sure if this is critical, but at least for Japanese (#321) and German (https://github.com/rust-lang/surveys/pull/322#discussion_r1874039609, https://github.com/rust-lang/surveys/pull/322#discussion_r1874041717) the results will be skewed; might be something to consider during the evaluation. But fixing that now in the survey might cause more confusion if some of the answers were for the old (incorrect) options, and some are for the corrected options.

---

> As they will improve the translation for next year(s).

But it could be 'wasted' work if the question is thrown out / modified for the next year, right?

---

> We never modify the survey while it is running, unless there is really something critical going on. We should perhaps clarify this in the blog post next time.

Yes that would be good to clarify. Or what about asking for translations and feedback in a separate blog post a few weeks or months before the survey starts (unless I overlooked that). That might be more productive, and might hopefully also reduce the chances of incorrect / outdated translations when the survey starts.
What do you think?

---

Maybe the survey could also explicitly link here (https://github.com/rust-lang/surveys) at the end, asking for issues / ideas for improvements for the survey in the next year? In case that would be useful.

Kobzol: > But it could be 'wasted' work if the question is thrown out / modified for the next year, right?

In theory yes, but usually the survey stays like 95+% the same year between year, and we always want to keep the best/latest/most updated version in the GitHub repository.

> Yes that would be good to clarify. Or what about asking for translations and feedback in a separate blog post a few weeks or months before the survey starts (unless I overlooked that). That might be more productive, and might hopefully also reduce the chances of incorrect / outdated translations when the survey starts.
What do you think?

We ask for help with translations on Zulip. We got 1-2 people for each language this year, which seemed to be enough, but of course, it's a lot of text, so having more eyes on it always helps. I also wanted to ask for help on Reddit, but I did not find the time and energy to do that, I was very busy during this autumn. Maybe next year we could also try to ask on Reddit in advance.

Marcono1234: Thanks for the explanation, and also thanks for organizing the survey in the first place!

Issue #320, opened 2024-12-06
Stabilizing features: Inconsistent translation for "Would unblock my use-case"


Marcono1234:
Hello,
for the question about stabilizing features (in English "Which unimplemented (or nightly only) features are you looking for to be stabilized?"), the option "Would unblock my use-case" seems to be translated inconsistently, which could potentially skew the results:
- English & most other languages
    > Would unblock my use-case

    Side note: The question here is though how strong this "my" is, and whether this option also applies if for example you develop multiple Rust projects and only one of it is blocked. Not sure how strict the other languages have translated this "my use-case" vs. "a use", or similar.
- Chinese (`zh-cn`)
    > 如果稳定版本中本功能可用，则会解决想用Rust适用我的用例时遇到的障碍

    German (`de`)
    > Würde es mir erlauben, Rust zu nutzen

    At least the German translation says "Would allow me to use Rust", which sounds as if due to this missing feature the user is not using Rust at all.
- Japanese (`ja`)
    > 高優先度

    The translation seems to be "high priority", which is quite unspecific compared to the other ones. Users might choose this option despite the missing feature not completely blocking them.

Kobzol:
Thanks! The "High Priority" was an answer last year, perhaps it was not re-translated this year. If you could send PRs that improve the translation, that would be great! The translations are [here](https://github.com/rust-lang/surveys/tree/main/surveys/2024-annual-survey/translations).

Marcono1234:
Sorry, I can only suggest an improvement for German. Should I chose a translation for "unblock my use-case" there too?

Kobzol:
Yes. that was the intended meaning.

PR #322, opened 2024-12-06
Fix typos and issues in German 2024 translation

Marcono1234:
Fixes some typos and issues with the German translations. I have added review comments explaining some of the changes.

Additional issues which are not resolved here:
- [x] For "problems to productivity", there is the option "Implementing dynamic library plugins"
The current German translation is "Dynamische Bibliotheken/Plugins" (so libraries _or_ plugins), I assume this is wrong. But I am also not sure what exactly the English version is referring to.
- [x] The option "Racial or ethnic minority" is translated as German "Rasse oder ethnische Minderheit"
In German the term "Rasse" (race) is normally used for animals, and there only being one 'human race'. If the term "Rasse" is used for humans it is often done in a racist way. Not sure what a better alternative would be here, maybe German "Abstammung" (?) or just writing "Ethnische Minderheit" (ethnic minority). Maybe another native German speaker could provide some feedback.
- [x] The list of countries for "Where do you live?" (German "Wo wohnen Sie?") might be out of order
_(see discussion below)_
For example "Österreich" (Austria) comes last for some reason. Not sure if this affects the survey, and if country names are mapped incorrectly to the respective English name there now.
- [x] The list of languages for "In which language ..." (German "Welche Sprache(n) ...") has a different order
Not sure if this causes the language names to be mapped incorrectly to the respective English name.
_(see discussion below)_

### DIFF
--- a/surveys/2024-annual-survey/translations/de.md
+++ b/surveys/2024-annual-survey/translations/de.md
@@ -289,7 +289,7 @@ Type: select one
 - Ich weiß es nicht oder es interessiert mich nicht
 - Andere Antwort...
 
-### Welche nicht implementierten (oder nur Nightly) Funktionen möchten Sie stabilisiert sehen?
+### Welche nicht implementierten (oder nur Nightly) Funktionen wünschen Sie sich?
 
 Type: matrix
 
@@ -317,7 +317,7 @@ Rows:
 
 Columns:
 
-- Würde eine Blockade bei meinem Anwendungsfall lösen
+- Würde ein Hindernis bei meinem Anwendungsfall beseitigen
 - Würde meinen Code verbessern
 - Brauche ich nicht
 - Weiß nicht, was das ist
@@ -336,7 +336,7 @@ Rows:
 - Quelltext über mehrere crates verteilen (z.B. Orphan Rule)
 - Iteratoren manuell implementieren
 - Nicht genug in konstanten Funktionen tun können
-- Dynamische Bibliotheken/Plugins implementieren
+- Dynamische Bibliotheken Plugins implementieren
 - Mit anderen Sprachen zusammenarbeiten (z.B. C oder C++)
 - Strukturierte Nebenläufigkeit mit Async Code erreichen
 - Async-Laufzeit-agnostischen Code schreiben
@@ -641,7 +641,7 @@ Type: select all that apply
 - Nicht-binäres Geschlecht
 - Älter oder jünger als die durchschnittlichen Entwickler, die ich kenne
 - Politische Ansichten
-- Rasse oder ethnische Minderheit
+- Ethnische Herkunft
 - Religiöse Ansichten
 - Trans
 - Frau oder als Frau wahrgenommen
@@ -667,7 +667,7 @@ Type: select one
 - < 3 Jahre
 - < 5 Jahre
 - < 10 Jahre
-- \> 10 Jahre
+- > 10 Jahre
 
 ### Wo wohnen Sie?
 
--- a/surveys/2024-annual-survey/translations/de.md
+++ b/surveys/2024-annual-survey/translations/de.md
@@ -35,7 +35,7 @@ Type: select all that apply
 - Rust ist zu schwer zu lernen, oder Rust zu lernen würde zu viel Zeit brauchen
 - Community war unhöflich, abweisend, oder auf andere Art unsympathisch
 - Ich benutze lieber eine andere Programmiersprache
-- Ich kann Rust aufgrund externe Umstände nicht nutzen
+- Ich kann Rust aufgrund externer Umstände nicht nutzen
 - Ich hatte noch keine Zeit, Rust zu nutzen
 - Anderer Grund
 
@@ -66,18 +66,18 @@ Type: select one
 Type: select one
 
 - Ich versuche immer noch *aktiv*, Rust zu lernen
-- Im Jahr 2023
-- Im Jahr 2021 oder 2022
-- Im Jahr 2019 oder 2020
-- Im Jahr 2017 oder 2018
-- Im Jahr 2016 oder früher
+- Im Jahr 2024
+- Im Jahr 2022 oder 2023
+- Im Jahr 2020 oder 2021
+- Im Jahr 2018 oder 2019
+- Im Jahr 2017 oder früher
 
 ### Wenn Sie Lernmaterial über Rust konsumiert haben, welche Art von Material haben Sie konsumiert?
 
 Type: select all that apply
 
 - Bücher („The Rust Programming Language“, „Rust for Rustaceans“ usw.)
-- Online-Übungen (Rustlings, Rust by Example, usw.
+- Online-Übungen (Rustlings, Rust by Example, usw.)
 - Videos oder Livestreams
 - Blogs
 - Dokumentation
@@ -117,7 +117,7 @@ Type: select all that apply
 - iOS
 - Android
 - Embedded Systeme (mit Betriebssystem)
-- Embedded Systeme (bare metal)
+- Embedded Systeme (ohne Betriebssystem)
 - WebAssembly (im Browser)
 - WebAssembly (für andere Zielsysteme)
 - Explizit plattformunabhängig (z.B. ein Crate, das nicht mit dem Betriebssystem interagiert)
@@ -145,8 +145,8 @@ Type: select all that apply
 
 - Die derzeitige stabile Version
 - Die vorige stabile Version
-- Eine spezifische stabile Rust-Version gleich oder neuer als 1.67
-- Eine spezifische stabile Rust-Version vor 1.67
+- Eine spezifische stabile Rust-Version gleich oder neuer als 1.75
+- Eine spezifische stabile Rust-Version vor 1.75
 - Die Beta Version
 - Aktuelle Nightly Version
 - Eine spezifische Nightly-Version
@@ -163,7 +163,7 @@ Type: select all that apply
 - Zur Nutzung eines oder mehrerer Sprachfeatures die ich brauche
 - Ich mag es, Zugriff auf die neuesten Features zu haben
 - Um zu helfen, die Nightly Version auf Bugs zu testen
-- Um Design-Feedback zu nächtlichen Funktionen zu geben
+- Um Design-Feedback zu Nightly Funktionen zu geben
 - Um in CI zu testen
 - Eine Crate die ich verwende benötigt Nightly
 - Ein Programm das ich verwende benötigt Nightly
@@ -269,9 +269,9 @@ Type: matrix
 Rows:
 
 - Ich kann die *stabile* Compilerversion aktualisieren, ohne befürchten zu müssen, dass mein Code nicht kompiliert werden kann
-- Ich kann die *nächtliche* Compilerversion aktualisieren, ohne befürchten zu müssen, dass mein Code nicht kompiliert werden kann
+- Ich kann die *Nightly* Compilerversion aktualisieren, ohne befürchten zu müssen, dass mein Code nicht kompiliert werden kann
 - Das Upgrade auf eine neue *stabile* Compilerversion erfordert entweder keine Änderungen oder nur äußerst kleine und einfache Änderungen an meinem Code
-- Das Upgrade auf eine neue *nächtliche* Compilerversion erfordert entweder keine Änderungen oder nur äußerst kleine und einfache Änderungen an meinem Code
+- Das Upgrade auf eine neue *Nightly* Compilerversion erfordert entweder keine Änderungen oder nur äußerst kleine und einfache Änderungen an meinem Code
 
 Columns:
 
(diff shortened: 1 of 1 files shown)


surveys/2024-annual-survey/translations/de.md
```diff
@@ -66,18 +66,18 @@ Type: select one
...
-- Im Jahr 2023
-- Im Jahr 2021 oder 2022
-- Im Jahr 2019 oder 2020
-- Im Jahr 2017 oder 2018
-- Im Jahr 2016 oder früher
+- Im Jahr 2024
+- Im Jahr 2022 oder 2023
+- Im Jahr 2020 oder 2021
+- Im Jahr 2018 oder 2019
+- Im Jahr 2017 oder früher
```
Marcono1234, line 73: Looks like German was the only language where this had not been adjusted; hopefully it does not skew the results

surveys/2024-annual-survey/translations/de.md
```diff
@@ -117,7 +117,7 @@ Type: select all that apply
 - iOS
 - Android
 - Embedded Systeme (mit Betriebssystem)
-- Embedded Systeme (bare metal)
+- Embedded Systeme (ohne Betriebssystem)
```
Marcono1234, line 120: Further below "bare metal" was also translated as "ohne Betriebssystem" (English "no operating system")

surveys/2024-annual-survey/translations/de.md
```diff
@@ -145,8 +145,8 @@ Type: select all that apply
 
 - Die derzeitige stabile Version
 - Die vorige stabile Version
-- Eine spezifische stabile Rust-Version gleich oder neuer als 1.67
-- Eine spezifische stabile Rust-Version vor 1.67
+- Eine spezifische stabile Rust-Version gleich oder neuer als 1.75
+- Eine spezifische stabile Rust-Version vor 1.75
```
Marcono1234, line 149: Looks like German was the only language where this had not been adjusted; hopefully it does not skew the results

surveys/2024-annual-survey/translations/de.md
```diff
@@ -163,7 +163,7 @@ Type: select all that apply
 - Zur Nutzung eines oder mehrerer Sprachfeatures die ich brauche
 - Ich mag es, Zugriff auf die neuesten Features zu haben
 - Um zu helfen, die Nightly Version auf Bugs zu testen
-- Um Design-Feedback zu nächtlichen Funktionen zu geben
+- Um Design-Feedback zu Nightly Funktionen zu geben
```
Marcono1234, line 166: For consistency keep the English term here instead of translating "nightly"; this is also consistent with usage in previous questions a few lines above

surveys/2024-annual-survey/translations/de.md
```diff
@@ -289,7 +289,7 @@ Type: select one
 - Ich weiß es nicht oder es interessiert mich nicht
 - Andere Antwort...
 
-### Welche nicht implementierten oder (nur nächtlichen) Funktionen möchten Sie stabilisieren?
+### Welche nicht implementierten (oder nur Nightly) Funktionen möchten Sie stabilisiert sehen?
```
Marcono1234, line 292: Not completely sure if the phrase "[...] stabilisiert sehen" actually exists in German, but the meaning is more correct probably. Before it said "which features do you want to stabilize?" (active instead of passive voice).

surveys/2024-annual-survey/translations/de.md
```diff
@@ -311,13 +311,13 @@ Rows:
...
 - Enum Varianten Typen
-- Allokator trait, bessere OOM-Verwaltung
+- Allokator Trait, bessere OOM-Verwaltung
 - Stabile ABI
 - Portables SIMD
 
 Columns:
 
-- Würde es mir erlauben, Rust zu nutzen
+- Würde eine Blockade bei meinem Anwendungsfall lösen
```
Marcono1234, line 320: Part of the discussion in https://github.com/rust-lang/surveys/issues/320

Suggestions for better wording are welcome though.

surveys/2024-annual-survey/translations/de.md
```diff
@@ -333,20 +333,20 @@ Type: matrix
 Rows:
 
 - Logik für Tupel variabler Länge implementieren
-- Quelltext über mehrere crates verteilen (Orphan Rule)
+- Quelltext über mehrere crates verteilen (z.B. Orphan Rule)
```
Marcono1234, line 336: "z.B." means "for example"; in the English version this is also only an example

surveys/2024-annual-survey/translations/de.md
```diff
@@ -333,20 +333,20 @@ Type: matrix
...
-- Nicht genug in konstanten funktionen tun können
+- Nicht genug in konstanten Funktionen tun können
 - Dynamische Bibliotheken/Plugins implementieren
 - Mit anderen Sprachen zusammenarbeiten (z.B. C oder C++)
 - Strukturierte Nebenläufigkeit mit Async Code erreichen
 - Async-Laufzeit-agnostischen Code schreiben
 - Korrekten unsafe Code schreiben
 - Der Borrow Checker lehnt meinen validen Code ab
-- Schlechte Performance zur Programmlaufzeit
+- Langsame Performance zur Programmlaufzeit
```
Marcono1234, line 345: German "schlecht" just means "bad", so it could be misunderstood as high memory usage. The English version says "slow" (German "langsam").

surveys/2024-annual-survey/translations/de.md
```diff
@@ -333,20 +333,20 @@ Type: matrix
...
 - Langsame Kompilierung
 - Große Binärdateien der kompilierten Programme
-- Hoher Speicherplatzverbrauch (der Dateien im target-Verzeichnis)
-- Kompilierfehler (z.B. intener Kompilerfehler, auch genannt ICE, oder Fehlkompilationen)
+- Hoher Speicherplatzverbrauch (z.B. der Dateien im target-Verzeichnis)
```
Marcono1234, line 348: "z.B." means "for example"; in the English version this is also only an example

surveys/2024-annual-survey/translations/de.md
```diff
@@ -391,14 +391,14 @@ Columns:
 - Ich wusste nicht, dass dies stabilisiert wurde
 - Ich weiß nicht, was das ist
 
-### Wie kompilieren Sie Ihre Rust Projekte?
+### Wie bauen Sie Ihre Rust Projekte?
```
Marcono1234, line 394: English version also says "build" (German "bauen"), I guess that covers more than compilation, e.g. also generating documentation

surveys/2024-annual-survey/translations/de.md
```diff
@@ -546,7 +546,7 @@ Type: select all that apply
 - Programmiersprachen und zugehörige Programme (einschließlich Compiler, IDEs, Standardbibliotheken usw.)
 - Robotik
 - Wissenschaftliches und/oder numerisches Rechnen
-- Serverseitige oder &quot;Backend&quot;-Anwendung
+- Serverseitige oder "Backend"-Anwendung
```
Marcono1234, line 549: Note: This was not strictly needed, it looks like in the survey they were properly displayed as `"`

surveys/2024-annual-survey/translations/de.md
```diff
@@ -667,7 +667,7 @@ Type: select one
 - < 3 Jahre
 - < 5 Jahre
 - < 10 Jahre
-- > 10 Jahre
+- \> 10 Jahre
```
Marcono1234, line 670: This is currently rendered as Markdown block quote here on GitHub

(actually affects all translations?)

surveys/2024-annual-survey/translations/de.md
apiraino: this is unfortunate

apiraino:
>  The list of countries for "Where do you live?" (German "Wo wohnen Sie?") might be out of order

In German, letters with Umlaut (äöü+ß) are [at the end of the alphabet](https://de.wikipedia.org/wiki/Deutsches_Alphabet).

Marcono1234:
Ah good point, but how do you map the results between the different survey languages then? I had assumed the answers are stored by numeric index, but if the countries and languages have a different order depending on the translation that seems problematic.

Kobzol:
It's not exactly stored by numeric index explicitly, but it's true that we cannot reorder the answers in the translations. So we should just store the order in the Markdown to match the order on SH, to avoid having a diff between SH and MD. In other words, there's not much point in modifying the order of answers in the translation markdown files.

Marcono1234:
> > The list of countries for "Where do you live?" (German "Wo wohnen Sie?") might be out of order
> 
> In German, letters with Umlaut (äöü+ß) are [at the end of the alphabet](https://de.wikipedia.org/wiki/Deutsches_Alphabet).

Just had a look at that article again, and that only applies if they appear standalone. As part of words they are apparently sorted near the corresponding non-Umlauts, see <https://de.wikipedia.org/wiki/Alphabetische_Sortierung#Sortierregeln_nach_Sprachen>.
Could be a problem if users from Austria (German speaking) are unable to find their country name "Österreich" where they expect it.

(Though I don't know if that is an issue with the survey software, which you cannot control.)

surveys/2024-annual-survey/translations/de.md
apiraino: maybe it's me but I associate "Blockade" with political contexts

surveys/2024-annual-survey/translations/de.md
apiraino: The deciding factor should be which is the most common used term by the Rust German-speaking community. What would your experience in the German Rust communities suggest?

FWIW this unofficial translation of the Rust book[^0] says "nachtlich", so I am inclined to think it is commonly accepted? But I can't really say.

[^0]: https://rust-lang-de.github.io/rustbook-de/ch00-00-introduction.html

surveys/2024-annual-survey/translations/de.md
apiraino: wünschen?

apiraino:
>  For "problems to productivity", there is the option "Implementing dynamic library plugins"
The current German translation is "Dynamische Bibliotheken/Plugins" (so libraries or plugins), I assume this is wrong. But I am also not sure what exactly the English version is referring to.

Probably there's not a huge difference but that phrasing was suggested in #294 (see also link to Zulip). I think it's probably better "Dynamische Bibliotheken Plugins"? What do you think?

apiraino:
@Marcono1234 first of all thanks for your time to review the translation :slightly_smiling_face:  I left a few comments, when you have time (not urgent). Apart from that, I'm happy to merge.

surveys/2024-annual-survey/translations/de.md
Marcono1234: Yeah, maybe "Hindernis ... beseitigen" instead?

surveys/2024-annual-survey/translations/de.md
apiraino: Is this change needed? AFAICS the list of options seems to be correctly rendered (unless it was fixed on our platform when we inputted the data)

<details><summary>Screenshot from the survey</summary>
<p>

grafik


</p>
</details>

surveys/2024-annual-survey/translations/de.md
Marcono1234: I am not active in the German Rust community, sorry.

Though I guess the term "nightly" is likely familiar to German users, since it occurs in compiler errors and also in Cargo commands.

In the German translation of the old Rust website it was called "Nightly" as well: https://github.com/rust-lang/prev.rust-lang.org/blob/master/de-DE/documentation.md
(side note: the new website has no official German translation)

Also, when restricting Google search results to German, it seems "rust" / "rust lang" "nightly" gives more results whereas "nächtlich" seems to already give results about the German city Rust and the video game Rust on the first result page.

surveys/2024-annual-survey/translations/de.md
Marcono1234: Not strictly needed if you copy the text here from the code and not there rendered version.
The issue was more with Markdown rendering (e.g. here on GitHub) where `> ...` is shown as block quote:
Rendered Markdown screenshot

I will revert it again. If this change is desired (and does not complicate copying the text from the source), might be better to apply it to all translations then.

Marcono1234:
> I think it's probably better "Dynamische Bibliotheken Plugins"? What do you think?

Originally I also thought maybe "plugin" refers to a specific Rust / Cargo concept I am not aware of, instead of just the general concept of plugin functionality (however that is implemented). But that is not specific to the German translation.

I have applied all your suggested changes I think. Please let me know if you want anything else changed 🙂

surveys/2024-annual-survey/translations/de.md
apiraino: :man_shrugging: 

let's go with this

Issue #247, opened 2023-12-11
Tasks for when the survey will stop accepting answers


apiraino:
I'll draft a list of TODOs to do when the survey will stop accepting answers

- [ ] Analyze the results
- [ ] Publish a blog post with the results
- [ ] Add a link to the announcement blog post pointing to the results blog post
- [ ] Update [forge.rust-lang.org](https://forge.rust-lang.org/community/survey-faq.html#where-can-i-see-the-previous-survey-reports) pointing to blog post with the results
- [ ] Update [FAQ](https://github.com/rust-lang/surveys/blob/main/documents/Community-Survey-FAQ.md)
- [x] Answer [this question](https://github.com/rust-lang/surveys/issues/9) about how answer datasets are stored

(please feel free to add/edit items)

PR #328, opened 2025-01-16
Add a guide for what to do after the State of Rust survey finishes

Kobzol:
This PR subsumes https://github.com/rust-lang/surveys/issues/247, which can be closed afterwards. It's better to store this documentation in the repo, rather than in an issue.


apiraino:
The compile steps described here will likely need some tweaking, but that can happen in another pass. Thanks!

PR #334, opened 2025-02-25
Add pdf skip script param

apiraino:
I've added a param to the `main.py` script to skip the PDF generation and reworded the README.

wdyt @Kobzol ?

thanks


Kobzol (approved):
I usually just comment it in the source code :laughing: But having an argument for that is better. Thanks!

PR #338, opened 2025-06-07
2025 council survey

Mark-Simulacrum:
Initial draft... needs some more work, or perhaps just cutting FIXMEs and publishing. Very open to feedback on the new approach.

cc https://rust-lang.zulipchat.com/#narrow/channel/392734-council/topic/Performance.20feedback.202025

cc @rust-lang/leadership-council

### DIFF
--- a/surveys/micro/leadership-council-2025.md
+++ b/surveys/micro/leadership-council-2025.md
@@ -39,6 +39,28 @@ See also our [spending/budget](https://github.com/rust-lang/leadership-council/b
 
 [duties]: https://rust-lang.github.io/rfcs/3392-leadership-council.html#duties-expectations-and-constraints-on-the-council
 
+### Do you consider yourself a member of the Rust Project?
+
+Type: select one (required)
+
+- Yes
+- No
+
+### What teams/areas of the project do you regularly participate in?
+
+This question aims to provide us with feedback on areas of the project that we
+may not be addressing as well (or areas we're doing better in).
+
+Type: select one or more (optional)
+
+- Compiler
+- Dev Tools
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Hm, we should probably snapshot that somewhere that's not world-writable if we want to link to it -- not sure we do, but it seems possibly reasonable? I guess I can put it in a gist or so... maybe we should keep it in a GitHub issue description?

ehuss:
Hm, yea. I changed write permissions to Owners/Admins, but snapshotting it somewhere sounds reasonable. A gist or somewhere on GitHub seems good.

Mark-Simulacrum:
OK, updated with the update post links and put up https://github.com/rust-lang/leadership-council/pull/189 to snapshot the budget (and linked that too).

It might be good to update the budget snapshot with recent decisions (e.g., continuing ops work is not reflected I think?)

ehuss:
I updated the hackmd with the compiler ops update. I hadn't added it before because it hasn't technically been accepted since it is still in FCP. But I don't expect any concerns blocking it.

surveys/micro/leadership-council-2025.md
```diff
@@ -0,0 +1,159 @@
...
+- Dev Tools
+- Launching Pad
+- Library
+- Infrastructure
+- Moderation
+- Language
+
+### Tracking and delegating work
+
+> Identifying, prioritizing, and tracking work that goes undone due to lack of
```
Kobzol, line 66: Is this supposed to be an internal comment or should this be shown in the survey?

PR #348, opened 2025-09-22
Add question about error codes to the annual survey

Kobzol:
Discussed in https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Error.20codes.20still.20useful.3F/with/540785396.


shepmaster:
Just a note that error codes are exposed in more ways than just `--explain`. For example, the playground extracts them and links to the online error index, and editors allow opening the error index locally (at least whatever emacs setup I have does)

programmerjake:
yeah, I think there should be another option for "I don't use --explain but error codes are useful for other things" or something.

PR #361, opened 2025-10-16
Finalize Ukrainian translation of annual survey 2025

tyranron:
Following the #359, this PR makes the final adjustments for Ukrainian translation of annual survey 2025, as well as fully synchronizes it with the version on SurveyHero (by the link provided to me by @apiraino).

At this point, the version on SurveyHero is fully up-to-date and is ready to go.


tyranron:
@Kobzol yeah, I've seen that in docs. It's just a matter of habit. I feel much more comfortable to work with text in IDE + VCS, and then export it wherever is needed.

apiraino:
@tyranron can we merge this translation so we can check again against the SurveyHero content?

PR #365, opened 2025-10-25
Add support for ranking questions on SurveyHero

Kobzol:
I used this question type in the Rust contributor survey. By the way, since that survey uses descriptiosn a lot, I also found some mismatches between SH/MD descriptions, so https://github.com/rust-lang/surveys/pull/363 should be really useful :)

### DIFF
--- a/verifier/src/api.rs
+++ b/verifier/src/api.rs
@@ -75,7 +75,10 @@ impl Client {
             .send()?
             .error_for_status()?;
 
-        let elements: Elements = response.json()?;
+        let text = response.text()?;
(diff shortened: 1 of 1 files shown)


apiraino:
Thanks!

I'll have a closer look next week but wanted to mention that while working on #363 at some point I've realized that we should probably switch to using the [`/elements` endpoint](https://developer.surveyhero.com/api/#element-api) instead of the [`/questions` endpoint](https://developer.surveyhero.com/api/#get-only-questions-from-survey) to get some missing bits of the survey.

I *think* this PR complements this future work and doesn't need to be reworked (I do hope well that their `Question` structure is the same across different endpoints!).

PR #369, opened 2025-11-10
Set a more specific binary name for the verifier tool

apiraino:
I tend to compile the tooling I use for the Rust project instead of running `cargo run`, I am not fond of giving GB of space away for artifacts.

This little patch change the name of binary artifact of the verifier tool from `verifier` (uh) to `rust-survey-verifier`.

(I'm fine with any other name as long as it's mnemonic, not too generic and less prone to collisions).

p.s. Could probably be an idea establishing a policy for binaries for tools we use internally

### DIFF
--- a/verifier/Cargo.toml
+++ b/verifier/Cargo.toml
@@ -3,6 +3,9 @@ name = "verifier"
 version = "0.1.0"
 edition = "2018"
 
+[[bin]]
+name = "rust-survey-verifier"
+
 [dependencies]
 anyhow = "1"
 clap = { version = "4.5", features = ["derive"] }


Kobzol (approved):
I find the `rust` prefix a bit redundant, but since currently there are exactly two people using this binary (you and me), it doesn't really matter much.

apiraino:
wait a sec this patch is wrong, right? I can't just rename the output file of the compilation this way?

Kobzol:
I thought it would rename the default binary if `path` is not set, but that's not how it works, apparently.

apiraino:
I was also fooled by that, I was SO SURE I didn't even test :sweat_smile: , I'm sorry
```
$ cargo build 
error: failed to parse manifest at `/path/rust/surveys/verifier/Cargo.toml`

Caused by:
  can't find `rust-survey-verifier` bin at `src/bin/rust-survey-verifier.rs` or `src/bin/rust-survey-verifier/main.rs`. Please specify bin.path if you want to use a non-default path.
```

PR #371, opened 2025-11-11
Remove 'zero width space' from survey translations

Marcono1234:
I assume they were added accidentally?

(Have a look at the diff in a local Git client or your IDE to better see the whitespace changes)


apiraino:
I don't see these whitespaces on SurveyHero (or they just are just elided from the GUI). I assume they are either there but not visible or are artifacts from the export t omarkdown.

I'll self-assign for investigation

Thanks for reporting this :)

@rustbot claim

apiraino:
old habits are hard to die

PR #406, opened 2026-03-31
add infra 2026 survey

marcoieni:
We would like to send this survey to the Rust Project members.

### DIFF
--- a/surveys/2026/infrastructure-feedback-survey/questions.md
+++ b/surveys/2026/infrastructure-feedback-survey/questions.md
@@ -2,7 +2,7 @@
 
 This survey is organized by the Rust Infrastructure Team.
 
-Its goal is to help the Rust Infrastructure team understand what it should focus on next.
+Its goal is to help the Rust Infrastructure Team understand what to focus on next.
 
 Please, answer only if you are a Rust Project member.
 
@@ -21,7 +21,7 @@ Type: select one (optional)
 - 3
 - 4
 
-### How satisfied are you with the support provided by the Rust Infra team?
+### How satisfied are you with the support provided by the Rust Infrastructure Team?
 
 (1 = Very Dissatisfied, 4 = Very Satisfied)
 
--- a/surveys/2026/infrastructure-feedback-survey/questions.md
+++ b/surveys/2026/infrastructure-feedback-survey/questions.md
@@ -6,65 +6,9 @@ Its goal is to help the Rust Infrastructure team understand what it should focus
 
 Please, answer only if you are a Rust Project member.
 
-### Your GitHub username
+The survey is anonymous, though you can optionally provide information about your GitHub username and Rust teams you are a part of.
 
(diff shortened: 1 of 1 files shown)


Kobzol:
Made some small tweaks to bring the survey closer to the format we usually use.

marcoieni:
I like your changes 👍

marcoieni:
Can we make the matrix single choice?
image

Kobzol:
Good catch, done.

marcoieni:
When selecting a rust team, "Other (open response)" doesn't work
image

Kobzol:
Fixed (the automatic exporter from MD to SurveyHero clearly has some bugs :) but it still saves time a lot).

surveys/2026/infrastructure-feedback-survey/questions.md
```diff
@@ -0,0 +1,137 @@
+# Survey questions
+
+This survey is organized by the Rust Infrastructure Team.
```
apiraino, line 3: This survey seems to assume that the infrastructure is just our business and does not mention Github in any way.

For example, I am frequently rate-limited when I use the triagebot but I know T-infra can't do much about it. I wonder if this survey should hint at this split in responsabilities anywhere or maybe have a question about "what would you like GH to do for T-infra", "do you see GH shortcomings affecting our infra?" or something like that.

Unsure how to formulate this, though :)

Kobzol: That's a good point, though I wonder if we should maybe do a completely separate survey just about GitHub annoyances? :laughing: As there might be a lot to unpack there.

But we can also add a question about GH troubles specifically to this survey. @marcoieni What do you think?

marcoieni: We left GitHub and all the others providers we use such as AWS, Datadog, etc out of the survey because we wanted to focus on things we could improve this year.

Also, there are items like "Reliability of CI" that regard GitHub directly and plenty of long text input boxes where people can talk about their pain points or suggestions, so people are free to mention whatever they want there 👍

If we include questions like

> "what would you like GH to do for T-infra", "do you see GH shortcomings affecting our infra?" 

I'm worried people would spend all their energy replying to those and that would be not actionable for us. Except for telling GitHub for the Nth time that people want a more reliable platform. But I'm sure GitHub is aware of this.

Instead, what I would like to take from this survey is stuff like "I have this pain point when interacting with bors". This is actionable and we can improve it!

I agree with Jakub that if necessary, we should do a separate survey just about GitHub and report the results to GitHub.

Kobzol: Ok, makes sense to me!

apiraino (approved):
@marcoieni since the survey is now running, can we merge this? Anything you wanted to add?

Kobzol:
As it's running, we can merge.

PR #407, opened 2026-04-03
feat: Add the 2026 Debugging survey report

hashcatHitman:
This is a first draft. I think it's pretty good so far, but there are questions that need answering still. Please see the comments starting with `TODO:`.


surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,467 @@
...
+    normalized = defaultdict(int)
+    for answer in answers:
+        answer = answer.strip().lower()
+        normalized[answer] += 1
+    items = sorted(normalized.items(), key=lambda x: x[1], reverse=True)
+    for value, count in items:
+        print(f"{value}: {count}")
+
+
+# TODO: The original copied version of this function assumed any `Question.kind`
```
Kobzol: The code doesn't really ignore MatrixQuestion, I think, it just assumes that we use the correct question type for the correct chart (e.g. a simple question for a bar chart). There are definitely asserts in the glue code that check that the correct question type was passed though.

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,467 @@
...
+    )
+    db = parse_surveyhero_answers(
+        Path(ROOT_DIR / "data/2026/debugging-responses.csv"),
+        year=2026,
+        summary=summary,
+    )
+
+    report = ChartReport()
+
+    # TODO: Should this be a bar chart instead? One of the responses makes up a
```
Kobzol: I think it is clear from context what is the fourth section, but bar chart is fine too. Up to you.

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,467 @@
...
+            raise NotImplementedError(
+                f"assert_same doesn't know how to handle the QuestionKind {type(q_summary.kind)}"
+            )
+
+    assert q_summary.is_single_answer() == q_answers.is_single_answer()
+
+
+# TODO: including_secret_data is currently unused. Figure out what information
+# qualifies as "secret data" and decide if it can be removed.
+def analyze(including_secret_data: bool) -> ChartReport:
```
Kobzol: We don't really have secret data here, this was just relevant for the contributor survey specifically.

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,467 @@
...
+        db.q_simple_single(debugger_user_past),
+    )
+
+    quit_because_of_debuggers = "Were issues with debugging support the primary reason why you stopped using Rust?"
+    report.add_pie_chart(
+        "were-issues-with-debugging-support-the-primary-reason-why-you-stopped-using-rust",
+        db.q_simple_single(quit_because_of_debuggers),
+    )
+
+    other_languages_diff = {
```
Kobzol, line 178: It would be cool to see the dependency of "I use debuggers in other languages" on "I use debuggers on Rust". But it's not exactly trivial to do this though (looks something like this: https://github.com/rust-lang/surveys/blob/94953c0194fd955a6f8a46968a0c4741b94149b2/surveys/2025/contributor-survey/report.py#L227).

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,467 @@
...
+    }
+    other_languages = "Do you use debuggers in other programming languages?"
+    report.add_pie_chart(
+        "do-you-use-debuggers-in-other-programming-languages",
+        db.q_simple_single(other_languages).rename_answers(
+            other_languages_diff  # ty:ignore[invalid-argument-type] TODO: I think this is a mistake in the type signature of `SimpleQuestion.rename_answers`.
+        ),
+    )
+
+    # TODO: I find this chart difficult to interpret. I think this should either
```
Kobzol: Yeah, this chart is misleading as-is. Things I think we should show here:
1. Absolute numbers as a bar chart, to show the overall popularity of individual tools
2. Popularity of these tools per OS. Ideally, this would be combined with 1) in a single stacked bar chart, where each bar would be split into OSes, like this: https://plotly.com/python/bar-charts/#bar-charts-with-long-format-data
3. A chart of the 10 most popular tool + OS combinations

Maybe also a stacked bar chart per OS, where the bars will be split by tool, but since there are many tools, that might be hard to read. Not sure if it's needed though.

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,467 @@
...
+        # read the percentage for (WinDbg/WSL). Can I make them all fairly easy
+        # to read somehow?
+        # textposition="inside",
+        legend_params=dict(
+            orientation="h",
+            y=-0.05,
+        ),
+    )
+
+    # TODO: Some of the most prominent words produced in the wordcloud are very
```
Kobzol: Tbh the wordclouds are just not very useful 🤷 The best way to describe the open answers is to just.. read them and then extract some shared sentiment from them into the blog post.

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,467 @@
...
+        "what-other-debuggers-or-workflows-do-you-use-wordcloud",
+        db.open_answers(other_debuggers),
+    )
+
+    debugger_used_for_diff = {
+        "Getting stack traces from hung/crashed processes": "Stack traces",
+    }
+    debugger_used_for = "What are you using debuggers for?"
+    report.add_bar_chart(
+        "what-are-you-using-debuggers-for",
```
Kobzol, line 308: Would be nice to show a chart of this separated by the used tool(s). E.g. GDB, ...

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,467 @@
...
+        ),
+        xaxis_tickangle=45,
+    )
+    debugger_used_for_responses = db.open_answers(debugger_used_for)
+    with open(
+        Path(__file__).parent / "open-response-debugger-used-for.txt", "w"
+    ) as f:
+        for answer in debugger_used_for_responses:
+            f.write(f"{answer}\n---\n\n")
+    # TODO: Similarly to the previous wordcloud related TODO, does it really
```
Kobzol: No, not really, but we can include the wordcloud if we already have it 😆

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,467 @@
...
+        xaxis_tickangle=45,
+    )
+    multilingual_responses = db.open_answers(multilingual)
+    with open(
+        Path(__file__).parent / "open-response-multilingual.txt", "w"
+    ) as f:
+        for answer in multilingual_responses:
+            f.write(f"{answer}\n---\n\n")
+    report.add_wordcloud(
+        "do-you-debug-programs-that-combine-rust-with-any-of-the-following-languages-wordcloud",
```
Kobzol, line 490: Assembly seems quite prominent here, we can count how often it appeared and add it as a virtual answer to the previous bar chart (see the `add_open` method).

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,467 @@
...
+        db.open_answers(debugger_difficulties),
+    )
+
+    step_through_issues_bool = "Do you experience any issues when trying to step through code with your debugger?"
+    report.add_pie_chart(
+        "do-you-experience-any-issues-when-trying-to-step-through-code-with-your-debugger",
+        db.q_simple_single(step_through_issues_bool),
+    )
+
+    # TODO: Could these be relabeled without the leading "When" and trailing
```
Kobzol: Maybe, but I don't think the answers are that long that we'd have to do this.

Kobzol (commented):
Thank you very much for working on this! I left a few comments.

It's a good start definitely, and I know that it takes a lot of time to go through the questions and prepare the charts. As a follow-up, I think we can add a few more charts, maybe even some cross-correlated charts, as suggested, and then of course write a blog post. Are you up for it? :)

surveys/2026/debugging/report/main.py
hashcatHitman: I'm not even sure `assert_same` has ever actually been used, now that I look for it.

surveys/2026/debugging/report/main.py
hashcatHitman: Yeah, I think I'll keep it as a pie chart.

surveys/2026/debugging/report/main.py
hashcatHitman: I added three new charts. I believe they cover everything you mentioned here. Let me know what you think?

surveys/2026/debugging/report/main.py
hashcatHitman: Wouldn't it be 0? The "use debuggers in other languages" question was only offered to those who either (have never used Rust) or (have previously used Rust but never used debuggers in Rust).

Kobzol: Nevermind, you're right; the thing I wanted to show is actually already in this chart - the % of people who **don't** use debuggers in Rust, but use them in other languages.

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,642 @@
...
+    )
+
+    # TODO: I find this chart difficult to interpret. I think this should either
+    # be replaced or accompanied by a bar chart or some other chart, possibly
+    # multiple. I want to better represent information such as:
+    # - Of all responses, X% of users use the specific combo "Debugger" + "OS"
+    # - Of all responses for "Debugger", X% of users did so on "OS"
+    # - Of all responses for "Debugger", X% of users selected more than one "OS"
+    # - Of all responses for "OS", X% of users used "Debugger"
+    # - Of all responses for "OS", X% of users selected more than one "Debugger"
```
apiraino: Possibly all these answers can be better explained in the blog post. These charts are slightly unwieldly and not very flexible for nuances. I am not sure but I would probably not think too much about the graph representation here.

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,642 @@
...
+    std_lib_pain_responses = db.open_answers_raw(std_lib_pain)
+    with open(OPEN_RESPONSES_DIR / "std-lib-pain.txt", "w") as f:
+        for answer in std_lib_pain_responses:
+            f.write(f"{answer}\n---\n\n")
+    report.add_wordcloud(
+        "what-standard-library-types-are-hard-to-work-with-when-debugging-wordcloud",
+        db.open_answers(std_lib_pain),
+    )
+
+    visualizer_attribute_known = "If you are a library author, are you aware of and using the debugger visualizer attribute?"
```
apiraino, line 559: The replies about the debugger visualizer are interesting, many don't know this tool (me included!); it could be useful to explain that in the blog post

surveys/2026/debugging/report/main.py
```diff
@@ -0,0 +1,642 @@
...
+    ) as f:
+        for answer in visualizer_attribute_avoided_responses:
+            f.write(f"{answer}\n---\n\n")
+    report.add_wordcloud(
+        "why-dont-you-use-the-debugger-visualizer-attribute-wordcloud",
+        db.open_answers(visualizer_attribute_avoided),
+    )
+
+    # TODO: Figure out shorter labels that are easier to read without losing
+    # significant information.
```
apiraino: If it helps, possible suggestions to shorten just the longest sentences:
```diff
-Suboptimal representations of values (e.g. Vec<T> is shown as a pointer, length and capacity rather than the elements)
+Bad values repr. (e.g. Vec<T> is shown as a pointer)
```

```diff
-Expression parser doesn't support expressions that you want to write
+Incomplete support for the Expression parser
```
```diff
-Incorrect information displayed (i.e. line numbers, variable values)
+Incorrect information displayed (i.e. line nums, var values)"
```

```diff
-Too much detail/irrelevant information (i.e. assembly views)
+Irrelevant details (i.e. assembly views)
```

The other seems readable and aren't longer than 2 lines

apiraino:
Nice work, thanks @hashcatHitman, I see a couple of interesting findings.

One additional general observation after looking at the PDF: I feel that these wordclouds don't add a great deal of meaningful info. But if you find them interesting, it's fine to leave them there.

surveys/2026/debugging/report/main.py
hashcatHitman: The charts I added have since been removed and replaced with your own, so I assume this is considered resolved?

surveys/2026/debugging/report/main.py
hashcatHitman: I added two charts for this in 8257fd5a2e6ef30b910d1f6acd1b43eb1f6f80c4. Let me know what you think?

surveys/2026/debugging/report/main.py
hashcatHitman: I adopted these shorter labels in 9f7a9c24da702cff539e0c405f8a773f1789faf8, though I had to further shorten "Incorrect information displayed" as it was still slightly overlapping with another label.

surveys/2026/debugging/report/main.py
Kobzol: Yes, I think that presenting it in the facet grids is easier to understand, thanks.

Kobzol:
If we decide to use the `What are you using debuggers for?` in the blog post, we would have to carefully explain how these were created, as it's quite complex :laughing: Thank you for creating them though.

Thanks a lot for polishing the charts! I think that we have interesting data there, and it's time to move on to the blog post. Feel free to chat with me on Zulip to discuss it :) I think that it might be nice to structure it in a similar way as the previous similar survey that we did about [compiler performance](https://blog.rust-lang.org/2025/09/10/rust-compiler-performance-survey-2025-results/), where we split the conversation into several subtopics that were indexed at the top of the post.

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+- [Closing Remarks](#closing-remarks)
+
+## Cohort Analysis
+
+The first step to making sense of the survey results is understanding who took
+the survey. We asked respondents to rate their Rust expertise, from "Never used
+it" to "Advanced". Over 80% reported themselves as "Advanced" or "Intermediate",
+split roughly evenly between the two:
+
+<!-- chart: how-would-you-rate-your-rust-expertise -->
```
Kobzol: I feel like these charts could become smaller, they don't contain that much information, but are currently pretty huge. `height=600` seems to do the trick.

```suggestion
<!-- chart: how-would-you-rate-your-rust-expertise (height=600) -->
```

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+
+We also asked respondents if they currently use or have used debuggers in Rust.
+Over 46% said they currently do, with the remaining responses split between
+"have in the past" and "never have":
```
Kobzol:
```suggestion
We also asked respondents if they currently use or have used debuggers in Rust.
Over 46% said they currently do, with the remaining responses split between
"have in the past" and "never have". That means that over half of respondents do not currently use a debugger for Rust!
```

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+
+For respondents who indicated they had previously used Rust but no longer did,
+we asked if challenges with debugging support were why they stopped. For nearly
+3%, the answer was "yes", with an additional 24% reporting debugging issues as
+being partially responsible (though mind the small response count; most
+respondents were active users of Rust):
+
+<!-- chart: were-issues-with-debugging-support-the-primary-reason-why-you-stopped-using-rust -->
+
+## Debugger Usage
```
Kobzol: I feel like this section should contain the "Do you use a debugger" chart? It seems kinda obvious that this question should belong here. Maybe just keep the cohort analysis and why people stopped using debuggers in the first section?

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+Over 46% said they currently do, with the remaining responses split between
+"have in the past" and "never have":
+
+<!-- chart: do-you-use-debuggers-in-rust -->
+
+Categorized by expertise, the responses reveal that roughly half of "beginners"
+have never used debuggers in Rust, and nearly half of "advanced users" currently
+do use debuggers in Rust:
+
+<!-- chart: do-you-use-debuggers-in-rust-per-expertise -->
```
Kobzol, line 51: This chart is a bit clipped:

Image

Adding some left margin should help.

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+
+We also asked respondents if they currently use or have used debuggers in Rust.
+Over 46% said they currently do, with the remaining responses split between
+"have in the past" and "never have":
+
+<!-- chart: do-you-use-debuggers-in-rust -->
+
+Categorized by expertise, the responses reveal that roughly half of "beginners"
+have never used debuggers in Rust, and nearly half of "advanced users" currently
+do use debuggers in Rust:
```
Kobzol:
```suggestion
Categorized by expertise, the responses reveal that roughly half of "beginners"
have never used debuggers in Rust, and advanced users are more likely to use a debugger in Rust:
```

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+<!-- chart: were-issues-with-debugging-support-the-primary-reason-why-you-stopped-using-rust -->
+
+## Debugger Usage
+
+Knowing what debuggers developers are using and how is another important part of
+understanding the challenges they face. To this end, we asked respondents how
+they were debugging their programs on different operating systems. Independent
+of operating system, most developers make use of print debugging and the `dbg!`
+macro. Excluding those, using `lldb` inside an IDE was the most popular choice,
+followed by `gdb` on the command line:
```
Kobzol:
```suggestion
Knowing what debuggers developers are using and how is another important part of
understanding the challenges they face. To this end, we asked respondents how
they were debugging their programs. Unsurprisingly, most developers make use of print debugging and the `dbg!`
macro. Excluding those, using the `lldb` debugger inside an IDE was the most popular choice,
followed by `gdb` on the command line:
```

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+
+<!-- chart: what-tools-and-workflows-do-you-use-to-debug-rust-programs -->
+
+Including the operating system as part of the response, we can get a more
+detailed breakdown of these results from two different angles. The first angle
```
Kobzol:
```suggestion
We can get a more detailed breakdown of these results if we include the operating system on which the respondents use a given debugging approach. We examine this from two different angles. The first angle
```

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+being, "On operating system X, what percent of responses are using debugger
+Y?". Print debugging and the `dbg!` macro are consistently the top two yet
+again, but looking beyond that, things get more interesting. On Linux, using
+`gdb` on the command line was the most popular choice by a thin margin, but on
+Windows, Windows Subsystem for Linux (WSL), and macOS, `lldb` in an IDE was the
+top pick with a much wider lead over the respective runner-up for each compared
+to Linux (which had `lldb` in an IDE as the runner-up). On Windows, the three
```
Kobzol: This sentence is complicated to follow, I didn't understand the "for each" part.

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+WinDbg and the Visual Studio debugger, which are primarily used on Windows, and
+`lldb`, which is used more on macOS than Windows in an IDE and on the command
+line:
+
+<!-- chart: what-tools-and-workflows-do-you-use-to-debug-rust-programs-per-os-2 -->
+
+As for how people actually use their debugger of choice, the aggregate results
+are not particularly surprising. Roughly 87% of users are using debuggers for
+stepping line-by-line through programs and a little over half of users are using
+debuggers to obtain stack traces from hung/crashed processes:
```
Kobzol:
```suggestion
To the 6 respondents who use WinDbg on Linux: we wish you luck!
```

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+
+<!-- chart: what-are-you-using-debuggers-for -->
+
+If we break these results down by expertise, we can learn a bit more about usage
+patterns. As users become more experienced with Rust, their use of debuggers for
+learning purposes decreases, and they get more stack traces from crashed
+processes:
```
Kobzol, line 113:
```suggestion
Only a quarter of the respondents use a debugger to debug async code. That might be partially caused by the async Rust debugging experience being clumsy and incomplete.

If we break these results down by expertise, we can learn a bit more about usage
patterns. As users become more experienced with Rust, their use of debuggers for
learning purposes decreases, but they spend more time analyzing stack traces from crashed
processes:
```

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+
+<!-- chart: what-are-you-using-debuggers-for-per-expertise -->
+
+The final bit of insight into how Rustaceans use debuggers is if they are
+debugging programs that use Rust alongside other programming languages. For 44%
+of respondents, the answer is "yes":
```
Kobzol:
```suggestion
The final bit of insight into how Rustaceans use debuggers is if they are
debugging programs that use Rust alongside other programming languages. For 44%
of respondents, the answer is "yes", which is a pretty high number!
```

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+
+<!-- chart: do-you-debug-programs-that-combine-rust-with-other-languages -->
+
+As for which languages those are, C dominates the scene at a little over 70%,
+followed by C++ at about 43% and Python at about 20%:
+
+<!-- chart: do-you-debug-programs-that-combine-rust-with-any-of-the-following-languages -->
+
+## Pain Points
```
Kobzol:
```suggestion
## Challenges
```

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+
+Instead of diving right into asking, "what problems do you face when using
+debuggers?", or something to that effect, we first asked respondents why they
+decide against using debuggers whenever they do, including for reasons that
+aren't necessarily "problems with debuggers". This can help to put the severity
+of any issues into context; if most users report they are blocked by poor
+support for language features and very few report being blocked by "non-issues",
+it'd show the lack of support to be a very significant blocker for users.
```
Kobzol: I didn't understand the second half of this sentence - what are the "non-issues"? Could you maybe reword this paragraph a bit?

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+about 22%, which is slightly more than issues with external library types, at
+about 20%:
+
+<!-- chart: when-you-dont-use-a-debugger-why-dont-you -->
+
+As stepping through code was anticipated to be one of the most common uses for
+debuggers, we directly asked respondents if they faced any issues when doing so.
+A little over 51% of respondents said they did:
+
+<!-- chart: do-you-experience-any-issues-when-trying-to-step-through-code-with-your-debugger -->
```
Kobzol: This doesn't really need a chart, it's just two values, it takes space unnecessarily.

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+
+We also directly asked respondents which types in the standard library were hard
+to work with, if any. This was an open-response question, and as can be seen in
+the wordcloud, a few common complaints were with `enum`s and collections,
+particularly `std::collections::HashMap` and `std::vec::Vec`:
+
+<!-- TODO: This doesn't work! Wordclouds really weren't designed to exist without a
+non-wordcloud chart. -->
+
+`chart: what-standard-library-types-are-hard-to-work-with-when-debugging-wordcloud`
```
Kobzol: I don't think we need to show the wordcloud, we can just describe it with a sentence.

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+As the survey was being created, we interviewed a few community members and did
+some brainstorming to think of a handful of pain points people might experience
+when using debuggers with Rust. We asked respondents to indicate which, if any,
+they have encountered. At slightly over 74%, poor representation of values was
+the most common pain point by a decent margin, followed by being unable to print
+variables at just over 55%:
+
+<!-- chart: which-of-these-pain-points-have-you-experienced-using-a-debugger-with-rust -->
+
+## Library Authors
```
Kobzol:
```suggestion
## Debugger visualizers
```

surveys/2026/debugging/report/blog.md
```diff
@@ -0,0 +1,237 @@
...
+
+We asked respondents to indicate if they were library authors, and if so, if
+they were aware of and using the `debugger_visualizer` attribute. Nearly 62% of
+respondents indicated that they were library authors who were not aware of this
+attribute:
+
+<!-- TODO: Fix this chart in the script, then consider rewriting this section. -->
```
Kobzol, line 182: This chart is also kinda useless, the 62% number if enough.

Kobzol (commented):
Thank you! I did a first pass through *most* of the text and left some comments.

hashcatHitman:
I still need to look through and address a lot of the review, but one of the things I'm worried about at the moment is accessibility. I don't need a screen reader and don't have much experience using one, but so far I haven't been able to learn anything about our charts using the Orca screen reader other than the information I explicitly described in paragraphs. Is there something we can do to address that?

It might just be the fact that I'm not very familiar with the tool, but I feel like this might be a problem regardless.

surveys/2026/debugging/report/blog.md
hashcatHitman: I'm hesitant to make this change because I feel the new wording may be implying a relationship that isn't necessarily true (something along the lines of "advanced users have the knowledge needed to benefit from debuggers", when the data could also support something like "using debuggers helps users learn enough about their code to consider themselves advanced users").

That, and I also don't want to reduce the amount of information directly available in the text if I can help it due to my accessibility concerns.

surveys/2026/debugging/report/blog.md
hashcatHitman: Am I understanding correctly that this suggested line should come after the chart it describes and replace the description for the chart that follows?

I do see that you've also suggested a new description for that following chart in https://github.com/rust-lang/surveys/pull/407#discussion_r3803817726, which would come after the chart. I'm not particularly against the suggested changes, but I worry that the exact placement of them could be a bit confusing. We usually have at least a sentence before each chart that "introduces it" (we did so for every chart in [the compiler performance survey report blog post]), it'd be a little odd for the only lines preceding this chart to be talking about the chart that came before it.

Also, the usual "accessibility concern has me worried about providing less information about the chart in text form".

[the compiler performance survey report blog post]: https://blog.rust-lang.org/2025/09/10/rust-compiler-performance-survey-2025-results/#overall-satisfaction

surveys/2026/debugging/report/blog.md
hashcatHitman: Concerns bundled in https://github.com/rust-lang/surveys/pull/407#discussion_r3806697905

surveys/2026/debugging/report/blog.md
hashcatHitman: Really? Is that based on the chart as it currently exists (broken!) or as it should exist? Because it *should* also show the percent who meet the criteria "library authors who knew about the attribute but did not use it", but it currently doesn't.

That criteria is directly relevant to the next chart, which was only offered to those individuals.

I feel like the chart would be more useful to have if it correctly displayed the four distinct categories.

surveys/2026/debugging/report/blog.md
hashcatHitman: The way I'm looking at it, "do you use a debugger" is more of a "who" question than a "how" or "why" question, particularly in the context of a survey all about debuggers. This is also part of why this question was the third question in the "Cohort Questions" portion of the actual survey.

Does considering that angle change your mind? If not, I'm happy to move things around as you suggested. I don't think it makes a huge difference either way.

Also, perhaps you misspoke, but:

> Maybe just keep the cohort analysis and why people stopped using debuggers in the first section?

That last question in the cohort analysis section is currently if debuggers were the reason people stopped using Rust, not why they stopped using debuggers.

surveys/2026/debugging/report/blog.md
hashcatHitman: I will have to think a bit on how to reword this properly. For now, though, I will try to explain what I am going for.

> On Linux, using `gdb` on the command line was
the most popular choice by a thin margin,

With `gdb` CLI on Linux at 11.5% and `lldb` in an IDE on Linux at 11.1%, `gdb` only has a small lead over `lldb` (the thin margin).

> but on Windows, Windows Subsystem for
Linux (WSL), and macOS, `lldb` in an IDE was the top pick with a much wider lead
over the respective runner-up for each compared to Linux (which had `lldb` in an
IDE as the runner-up).

On Windows, `lldb` in an IDE leads with 13.3%, and is followed by the Visual Studio debugger at 7.27%.

On WSL, `lldb` in an IDE leads with 10.7%, and is followed by `gdb` CLI at 4.58%.

On macOS, `lldb` in an IDE leads with 15.5%, and is followed by `lldb` CLI at 8.03%.

| OS | "Top Pick" | "Runner-Up" | Difference |
|--------|--------|--------|--------|
| Linux | `gdb` CLI (11.5%) | `lldb` in IDE (11.1%) | ~0.4% |
| Windows | `lldb` in IDE (13.3%) | VS debugger (7.27%) | ~6.03% |
| WSL | `lldb` in IDE (10.7%) | `gdb` CLI (4.58%) | ~6.12% |
| macOS | `lldb` in IDE (15.5%) | `lldb` CLI (8.03%) | ~7.47% |

So when looking at the top two non-print-debugging debuggers per OS:

On Linux, even though the `gdb` CLI is preferred, the runner-up is not far behind. One could say that people could go "either way" between the two.

On Windows, WSL, and macOS, the preferred choice beat out the runner-up by at least 6%, making it a much more decisive victory. For all three, the preferred choice was `lldb` in an IDE, but they had different "runner-up" choices.

surveys/2026/debugging/report/blog.md
hashcatHitman: I will ponder the best way to reword this, but for now, I will try to explain what I was going for.

The way I see it, some of these responses are clear indicators of a problem (i.e. "language features aren't supported") and others are much more likely to be either "not a problem" or "not *our* problem" (i.e. "my code works"). I think you could probably argue any of the responses could go either way (especially because of how we worded them), but I also think many of them "lean" a certain way (i.e., "I don't know how" reads to me as leaning towards "not a problem" as the user just needs to take the time to learn how, but you could certainly argue that this is caused by poor documentation, a UI/UX issue, or something else that makes it more of "our problem").

If 90% of people report they decide against using a debugger because their code works and 30% report they decide against using a debugger because language features aren't supported, that's concerning, but not the end of the world.

If we flip those numbers, then not only are we saying that 90% of people decided against using a debugger because language features aren't supported, but we're also saying that *only* 30% of them have ever decided against using a debugger because their code worked. I feel like that would suggest something about the severity of the lack of language support.

I might be trying to create meaning where there is none, though. Maybe it'd be better to drop this part?

surveys/2026/debugging/report/blog.md
hashcatHitman: Removed. I left the text, does that seem better?

surveys/2026/debugging/report/blog.md
Kobzol: Yeah sorry, I meant Rust, not debuggers :)

I'm fine with keeping the current structure in that case, but I don't like the "Cohort analysis" section name very much. Maybe we can directly base the section names based on what you write, and have "Who uses debuggers?" and "How are debuggers used?" or something like that?

surveys/2026/debugging/report/blog.md
Kobzol: I just found the original text confusing, because it says "half of beginners do X and half of advanced users do Y", it switches both the target group and X -> Y in the same sentence, which just makes it confusing for me to read.

surveys/2026/debugging/report/blog.md
Kobzol: I think that this is a lot of nuance that I'm not sure if we even have to dig into. Maybe if it provided some super interesting insights, but I don't personally see this as being particularly interesting. Maybe we can just say that LLDB is popular across operating systems and people on Linux are much more likely to use a CLI than on other OSes?

surveys/2026/debugging/report/blog.md
Kobzol: Sorry, I messed up the GitHub diff, I just wanted to add this as a new sentence, not remove the previous paragraph (the GitHub UI does not make it particularly easy to add new stuff at a specific place with a proposed diff). If you wan to keep the sentence, place it wherever you see fit :)

surveys/2026/debugging/report/blog.md
Kobzol: I might have also placed this in the wrong place.

surveys/2026/debugging/report/blog.md
Kobzol: I like the way you think deeply about these things :) I think that mentioning the high-level split is fine ("problems caused by poor debugger support" vs "I wouldn't use a debugger even if it was perfect"), but the rest is too complicated, so I would drop it, yeah.

surveys/2026/debugging/report/blog.md
Kobzol: Oh, sure, if the chart shows more information, then feel free to keep it. I just think that it is pretty much useless to have a gigantic pie chart that shows two categories. 3+ is fine.

Kobzol:
> I still need to look through and address a lot of the review, but one of the things I'm worried about at the moment is accessibility. I don't need a screen reader and don't have much experience using one, but so far I haven't been able to learn anything about our charts using the Orca screen reader other than the information I explicitly described in paragraphs. Is there something we can do to address that?

Possibly (and kudos to you for thinking about it), but I'm not sure if we have the capacity for that. If you know how to make the post more accessible, then by all means do, but in general we usually do not go to great lengths to measure and consider accessibility, both because we don't have the knowledge, but more pragmatically because processing the surveys and writing the analysis blog posts is already a lot of work as-is, and most of it is done by volunteers, and we just can't expect to have everything perfect in all axes.

I guess that one way would be to extend the blog post text to essentially contain *all* information from the charts, but that would make it very difficult to read for most people, I think.

Based on the little of what I know about accessibility, maybe adding an `alt` tag to the automatically generated charts would help? We could provide the alt metadata in the Python script, next to each chart. But again, we didn't really do this for any of the previous posts, and this post is already kinda long overdue, so I'm not sure if this is the best place to start.

surveys/2026/debugging/report/blog.md
hashcatHitman: If you like those section names better, that's fine by me.

surveys/2026/debugging/report/blog.md
hashcatHitman: Hm, that is a good point. Would breaking it up a bit help, maybe? Something like...

> Categorized by expertise, the responses reveal that roughly half of "beginners" have never used debuggers in Rust! On the other hand, nearly half of "advanced users" currently do use debuggers in Rust:

surveys/2026/debugging/report/blog.md
hashcatHitman: I think saying that using `lldb` in an IDE is a pretty popular option across operating systems would be reasonable. I'm not sure if saying people on Linux are much more likely to use a CLI is ideal, though. I can definitely see where that idea would come from, but I'm not sure it's the right wording. I think the second idea is more accurately represented with something like:

> Those on Linux-based and "Other" operating systems were more willing to work from the command line than others.

But that still feels like getting ahead of ourselves a bit. I already pointed out the reluctance to use CLI on Windows and the popularity of "IDK" on Windows and macOS, so maybe it's fine to just drop this part and focus on the popularity of `lldb` in an IDE. Something like...

> On Linux, using `gdb` on the command line was the most popular choice by a thin margin, beating `lldb` in an IDE by only 0.4%. On Windows, Windows Subsystem for Linux (WSL), and macOS, `lldb` in an IDE was the top pick by at least 6%, making it a very popular choice in general.

surveys/2026/debugging/report/blog.md
hashcatHitman: Aaaah, gotcha. Added.

surveys/2026/debugging/report/blog.md
hashcatHitman: Does c855aa03f334030cfddef5ba57f0c604231ebfd4 look good?

surveys/2026/debugging/report/blog.md
hashcatHitman: Dropped in cde4081ca74a6fb4b87553259d2dd95aebc4fc6e, I didn't add anything new since I feel the existing text already points out the high-level split. Let me know what you think?

hashcatHitman:
> > (accessibility concerns)
> 
> Possibly (and kudos to you for thinking about it), but I'm not sure if we have the capacity for that. If you know how to make the post more accessible, then by all means do, but in general we usually do not go to great lengths to measure and consider accessibility, both because we don't have the knowledge, but more pragmatically because processing the surveys and writing the analysis blog posts is already a lot of work as-is, and most of it is done by volunteers, and we just can't expect to have everything perfect in all axes.

Fair enough. I'd have to look into it more to figure out the best way to handle it, so I'll put that on the back burner for now.

> I guess that one way would be to extend the blog post text to essentially contain _all_ information from the charts, but that would make it very difficult to read for most people, I think.

Yeah, I don't think that's feasible, particularly with the debugger/OS charts.

> Based on the little of what I know about accessibility, maybe adding an `alt` tag to the automatically generated charts would help? We could provide the alt metadata in the Python script, next to each chart. But again, we didn't really do this for any of the previous posts, and this post is already kinda long overdue, so I'm not sure if this is the best place to start.

There already are `alt` tags, technically. If users don't have JavaScript enabled (or some other thing is going on that would trigger `noscript` tags), the dynamic charts are replaced with their static PNG versions. The `alt` text is just the filename without the extension, though. But I digress; I can worry about this more in the future.

surveys/2026/debugging/report/blog.md
hashcatHitman: Yeah, it was showing as two categories because we were normalizing the answers to the question in one place but we weren't in another place, leading to the three answers ending in periods being unknown and treated as "Other". I managed to fix this with a030ce9a657a18a4d9f5661d98c47d299259013f, though I'm not 100% sure if that's the right long-term solution or not. It *seems* fine to me.

surveys/2026/debugging/report/blog.md
Kobzol: Looks good :+1:

surveys/2026/debugging/report/blog.md
Kobzol: :+1:

surveys/2026/debugging/report/blog.md
Kobzol: That looks good, feel free to use that.

apiraino:
@hashcatHitman if you need support/ideas about accessibility, try asking on Zulip: maybe someone can help. As Jakub says, we are really lacking expertise here

surveys/2026/debugging/report/blog.md
hashcatHitman: This can be marked resolved if you're happy with the placement I chose in 5029165cfd9d501b05253bd1247bd3dbd50813dc.

surveys/2026/debugging/report/main.py
hashcatHitman: I take it this can be marked resolved?

surveys/2026/debugging/report/main.py
hashcatHitman: Feel free to let me know if you feel the explanation in the blog post seems good to you.

surveys/2026/debugging/report/main.py
hashcatHitman: I'm marking this as resolved since these charts were replaced, but feel free to mark it as unresolved or leave a new review comment if you feel this is still relevant in some way.

hashcatHitman:
> @hashcatHitman if you need support/ideas about accessibility, try asking on Zulip: maybe someone can help. As Jakub says, we are really lacking expertise here

I probably will (even if we can't ship anything at this time). I'm still not very good at using Zulip, though (there are so many channels and threads... and I'm only subscribed to a few of those channels!); do you have any particular advice about the best place to ask such a question? I was thinking maybe making a thread in [#all/private](https://rust-lang.zulipchat.com/#narrow/channel/546943-all.2Fprivate), but maybe this is better suited as an open question for anyone on Zulip? Presumably going in a thread in [#general](https://rust-lang.zulipchat.com/#narrow/channel/122651-general)?

surveys/2026/debugging/report/blog.md
Kobzol: Yup.

Kobzol:
I think that general is a good fit. By the way, I think that the blog post is ready for a wider review. But that shouldn't be done in this PR.

Could you please take the generated blog post, together with the images, and open a PR against https://github.com/rust-lang/blog.rust-lang.org? There we will finalize the reviews, by also asking others to take a look. Any changes made there can then be backported into this PR.

Kobzol:
Regarding the past few commits:
- https://github.com/rust-lang/surveys/commit/535585fbbd6f2379851f8da065509e65ee588e45 I find `dict` to be easier to read when nested like this, as there are less strings involved. The performance is unmeasurable and we don't care about that here anyway, so that's not really needed here. But keep it, it's not super important.
- https://github.com/rust-lang/surveys/commit/8a264c4ee84c82505e592981afdd98543971185d I really wouldn't say that this is idiomatic, rather that it is harder to understand what's going on, the for cycle is as simple code as can be.

What linter do you use, btw?

hashcatHitman:
>     * [8a264c4](https://github.com/rust-lang/surveys/commit/8a264c4ee84c82505e592981afdd98543971185d) I really wouldn't say that this is idiomatic, rather that it is harder to understand what's going on, the for cycle is as simple code as can be.

I can revert that if you prefer. I'm not particularly well versed on what "idiomatic" Python looks like, so I tend to defer to my tooling on that if it seems reasonable to me.

> What linter do you use, btw?

I use [Ruff](https://docs.astral.sh/ruff/linter/), alongside the [ty](https://docs.astral.sh/ty/) type checker and language server.

Kobzol:
Oh, I didn't realize that ruff flags `dict` by default. That's weird to me.

hashcatHitman:
> Oh, I didn't realize that ruff flags `dict` by default. That's weird to me.

The only custom settings I have set for Ruff are here:

https://github.com/hashcatHitman/dotfiles/blob/main/.config/VSCodium/User/settings.json#L137-L138

hashcatHitman:
> Any changes made there can then be backported into this PR.

To get clarification on this: are you saying I should make changes there first, directly to the HTML? Or would it be fine and reasonable to make changes here first and just generate the changes that are to be made over there?

hashcatHitman:
> I think that general is a good fit. By the way, I think that the blog post is ready for a wider review. But that shouldn't be done in this PR.

> @hashcatHitman if you need support/ideas about accessibility, try asking on Zulip: maybe someone can help. As Jakub says, we are really lacking expertise here

[#general > Looking for advice on accessibility in survey blog posts](https://rust-lang.zulipchat.com/#narrow/channel/122651-general/topic/Looking.20for.20advice.20on.20accessibility.20in.20survey.20blog.20posts/with/617735950)

Kobzol:
> > Any changes made there can then be backported into this PR.
> 
> To get clarification on this: are you saying I should make changes there first, directly to the HTML? Or would it be fine and reasonable to make changes here first and just generate the changes that are to be made over there?

Up to you, whatever will be more pleasant for you. I think that I used to make changes to the surveys repo, then regenerate the final HTML and commit & push that.

Kobzol (approved):
Thank you very much! Awesome work.

PR #408, opened 2026-04-10
chore: Ignore various secrets

hashcatHitman:
While working on #407, I've noticed I have a lot of files showing up in my `git status` that I have no desire to check in. These are mostly files we'd like to avoid checking in at all, from what I can tell. I think we're not taking advantage of `.gitignore` to the degree we ought to. The proposed changes are to ignore:

- `/data/`, the directory that seems to be where raw survey response data goes by convention.
- `.env`, files generally meant for local configuration and secrets. We have a sample checked in which suggests that some of us do use `.env` files to store API tokens.

In addition, I'd also like to suggest we ignore any or all of:

- `open-response-*.txt`, the pattern I used in #407 to make it easier to read multiple open response answers. We already ignore `answers.txt`, but this pattern is more flexible. We could also consider `open-response-*.*` as an alternative, to account for the possibility of using a different format in the future, such as markdown.
- `*.secret.*`, a pattern I have used in my own projects to make it easy to ignore a file without needing to change my `.gitignore` or even reveal the existence of said file. This allows `foo.bar` to be ignored by renaming it to `foo.secret.bar`.

I expect these additional suggestions to be slightly more controversial, and have not yet made the changes. Pending discussion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ answers.txt
 __pycache__/
 /data/
 .env
+/open_responses/
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target/
 answers.txt
 __pycache__/
 /data/
+.env
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target/
 answers.txt
 __pycache__/
+/data/


apiraino:
In principle I am not opposed to that, if it makes your life easier. I don't personally use the `*.secret.*` pattern but I do use `.env` (and in general I don't store unencrypted secrets on the filesystem, which I hope is an accepted practice :sweat_smile: ).

> We have a sample checked
I don't think I see that (f.e. a `.env.sample`) or did I misunderstand this sentence?

About ignoring `open-response-*.txt` or other variants, ideally I'd like to see a directory tree where we create these files so we could just ignore that directory. But as a stopgap solution, ok for me as you suggest here.

Kobzol:
`/data` and `.env` sound fine. I don't use the `.secret` pattern, but adding a few entries to `.gitignore` doesn't really hurt, as long as it's not files that we would often commit, so I don't mind adding it.

hashcatHitman:
>> We have a sample checked

> I don't think I see that (f.e. a `.env.sample`) or did I misunderstand this sentence?

I am referring to this:
https://github.com/rust-lang/surveys/blob/main/surveyhero/.env.sample

> About ignoring `open-response-*.txt` or other variants, ideally I'd like to see a directory tree where we create these files so we could just ignore that directory. But as a stopgap solution, ok for me as you suggest here.

Do you think something like a top-level `open_responses` directory would be better? Or maybe it would be per-survey and go next to each `report` directory (i.e. `surveys/2026/debugging/open_responses`)?

apiraino:
> I am referring to this: [`main`/surveyhero/.env.sample](https://github.com/rust-lang/surveys/blob/main/surveyhero/.env.sample?rgh-link-date=2026-04-10T13%3A53%3A38Z)

Ah yes, *I* even added that file :)

> Do you think something like a top-level `open_responses` directory would be better? Or maybe it would be per-survey and go next to each `report` directory (i.e. `surveys/2026/debugging/open_responses`)?

I think I'd prefer a single top-level `open_reponses` directory tree as it seems the most practical solution, wdyt?

hashcatHitman:
> I think I'd prefer a single top-level `open_reponses` directory tree as it seems the most practical solution, wdyt?

I think that works out great; it's easy enough to add the top-level `open_responses` directory to the `.gitignore`, and it'd still cover it if someone wanted to use additional organization within (i.e. `open_responses/2026/debugging/foo.txt`).

hashcatHitman:
I'm opting to leave out the `*.secret.*` pattern since the reception was more "I don't mind" than "that seems useful". It was just an idea I thought *might* seem useful, but as it stands I don't currently need it either.

As far as I am concerned, this is ready to merge, but additional comments are welcome.

Kobzol:
Fine by me :) We can always change it.

apiraino:
thanks!

PR #409, opened 2026-06-14
Add 2026 Leadership Council review survey

Kobzol:
CC @rust-lang/leadership-council to take a look

The second commit contains changes vs the last year's survey.


apiraino:
One piece of feedback that might end up being slightly opaque is when asking for feedback on specific items taken from the LC description. These items are pretty wide and a bullet point feedback can provide a "vibecheck" but doesn't leave room to argument. Take for example: "Establishing the official position, opinion, or will of the Rust Project as a whole.": how can someone tick a box and confidently say that it represents their opinion?

If you think you can manage the expected amount of responses, maybe an optional text field could help addressing this.

Kobzol:
I think that in this survey the open answers will indeed be the most interesting thing.

Though I'm not necessarily a big fan of having 10 open answers scoped to the individual questions, I think that people generally tend to write one open-ended feedback into the final open answer, because it gets a bit tricky to split an open answer into 10 text fields if it spans multiple topics.

Do you consider the final open answer "anything else you want to add?" to be enough for this? We can clarify in the last open answer that the respondent could make clarifications on the previously mentioned duties of the council.

apiraino:
> We can clarify in the last open answer that the respondent could make clarifications on the previously mentioned duties of the council.

Maybe that can help.

I just think that the section "Leadership Council Effectiveness" cannot be evaluated with a bullet-point list in a satisfactory way, there's just too much implied in it. I am reading your mind, you discussed the set of questions, you are honestly interested in a feedback. I just don't want people to be led to thinking that the LC throws a survey with standard "corpo-like questions" because it's written in an RFC.

Kobzol:
I think that both open answers and the bullet-point list is valuable, because we want to both hear the nuanced opinions of Project Members, but also have a simple numerical way of evaluating their vibes. Of course, such numerical evaluation will always be somewhat misleading, but it still gives *some* signal, e.g. if 90% of respondents give us 1/5, then it's clear that something is wrong, and it's much faster and easier to understand this (and relay this to others) via a few numbers, than via many text responses.

It is also useful for explaining the results in a blog post, because we can't really share the open answers, and it is tricky to summarize them in an unbiased way. While *sharing* the numbers doesn't introduce any bias, and is a simple way to eyeball some commonly understood metrics.

Then the remaining question is whether we should have one general open answer at the end, or N open answers throughout the survey.

apiraino (approved):

lqd (approved):
(there's a tiny typo on the md filename itself :3)

lqd:
Let's talk about it at this week's LC meeting though (and Jakub has already nominated it). Now could be a good time to see if the previous iterations have yielded the actionable feedback the LC hoped when this mechanism was set up, if there could be improvements to make, and so on.

surveys/2026/leadership-council/questions.md
```diff
@@ -0,0 +1,187 @@
...
+
+This is a quick annual survey for Rust-lang project members to provide feedback
+on the Leadership Council. Your feedback is appreciated to help steer the
+Council activities. As always, if you have specific questions or feedback that
+you would like to provide, feel free to reach out to any Council member with
+questions.
+
+This survey is anonymous. The Council will go through all answers and release a
+summary after the survey is complete, similar to the one published in
+[past years](https://blog.rust-lang.org/2024/08/26/council-survey/).
```
Mark-Simulacrum, line 24: Hm, did we not run a 2025 survey? Or did the summary of results get lost somewhere?

Kobzol: The survey happened, but I didn't find any blog post.

Kobzol:
The survey has been launched.

PR #410, opened 2026-07-08
fix typo/translaction in zh-tw annual survey

hms5232:
* custom => 自訂
  * the original (自定義) is used in zh-CN
* local => 本機
  * the original has typo or redundant words


apiraino (approved):
thank you

